PR #33, opened 2019-07-11
Hotfix/erroneous ok response

msk-:
Probably good to have a review from both of you. If you merge, don't forget to squash.


src/index.js
```diff
@@ -242,6 +242,13 @@ const Errors = require('@modusbox/mojaloop-sdk-standard-components').Errors;
 
     // Handle requests
     inboundApi.use(router(inboundHandlers.map));
+    inboundApi.use(async (ctx, next) => {
+        // Override Koa's default behaviour of returning the status code as text in the body. If we
+        // haven't defined the body, we want it empty.
+        if (ctx.response.body === undefined) {
```
KamuelaFranco: What's the behavior if the body is `null` or something else that's falsy? Is a direct match for `undefined` the best approach?

msk-: Actually, I _should've_ set the body to `null`. In fact, I thought I did. Weird. `null` is Koa's idea of an empty body. So probably shouldn't overwrite in that case. The difference in response between the three is as follows:
- undefined: koa will overwrite with a text representation of the status code
- '': content-length: 0, content-type: text/plain
- null: no content response headers

msk-: (I've now made that change..)

PR #38, opened 2019-08-08
Cicd docker

dgethings:
This PR adds in the CICD config to create a docker image.

There are 2 scenarios an image is created:

- a push to a non-master branch
- a push of a git tag

The image tag is either the branch name (when a branch is non-master and not git tagged) or the git tag. If the push contains a git tag, that takes precedence.


dgethings:
Hold on. That's not right. The only change should be to add the .circleci/config.yml file. This PR isn't right. Closing PR and will issue a new one with the correct changes.

PR #43, opened 2019-08-20
Move package.json to root of the repo

ramirogm:
To be able to import the SDK from other projects via a github dependency ( at least ) the package.json needs to be on the root of the project.


bushjames:
@ramirogm - I am reluctant to change the scheme-adapter into a shared library in of itself. It is intended to be a stand alone container. My suggestion is to factor out those components you need to import into your own code as a shared library.

PR #45, opened 2019-08-27
Added ability to dynamically read JWS verification keys.

vbarzokas:


src/config.js
```diff
@@ -111,12 +124,26 @@ const setConfig = async cfg => {
...
         .map(f => {
             config.jwsVerificationKeys[path.basename(f, '.pem')] = fs.readFileSync(path.join(cfg.JWS_VERIFICATION_KEYS_DIRECTORY, f));
         });
 
+    // continuously monitor folder for changes in files.
+    fsWatcher = fs.watch(cfg.JWS_VERIFICATION_KEYS_DIRECTORY, (eventType, filename) => {
+        // On most platforms, 'rename' is emitted whenever a filename appears or disappears in the directory.
+        // From: https://nodejs.org/docs/latest/api/fs.html#fs_fs_watch_filename_options_listener
+        if (eventType === FS_EVENT_TYPES.RENAME) {
+            if (config.jwsVerificationKeys[path.basename(filename, '.pem')] == null) {
```
msk-: Unsure whether `===` beats `==` for comparison with `null` but perhaps worth using anyway.

src/config.js
```diff
@@ -125,13 +152,21 @@ const setConfig = async cfg => {
...
 };
 
-
 const getConfig = () => {
     return config;
 };
 
+// useful for closing the open handler of the fs.watch, especially in unit tests.
+const destroy = () => {
+    if (fsWatcher && typeof (fsWatcher.close) === 'function') {
```
msk-: Why do we test type of `fsWatcher.close` here?

bushjames:
Strange but it looks like the wrong circle ci is trying to test this.

src/config.js
vbarzokas: It will not work, as if you see two lines below that I use `delete` to completely remove that value, so strictly checking for `null` would return `false` as the value is actually  `undefined`.

This is the only case that I don't use triple `=`, as it is kind of safe since it is truthy only for `null` and `undefined` values. That is why eslint doesn't complain as well, while in other cases it would.

src/config.js
vbarzokas: Not a strict necessity, indeed. I just wanted to be sure that it is defined and it won't crash since the official documentation states for the method `close`:
```Stop watching for changes on the given fs.FSWatcher. Once stopped, the fs.FSWatcher object is no longer usable.```
And I had the impression that it won't even be defined after calling it, but now that I debug it I see that it is still there. Maybe I faced another issue during my tests, I am not sure. I can remove it.

src/config.js
msk-: Great, thanks

src/config.js
msk-: No harm leaving it, and it sounds like the docs aren't especially precise, so maybe we ought to leave it there. I just thought it was unusual.

PR #48, opened 2019-09-12
Feature/three stage transfers

bushjames:


MichaelJBRichards:
Looks great to me! Just what we wanted

PR #64, opened 2019-10-09
Implement expiry timestamp check for quotes and transfers requests.

kirgene:


elnyry:
@MichaelJBRichards @bushjames @kirgene - I've been skeptical about enforcing these at the Switch level because of a section I'm referencing below; I believe it should be left to the FSPs to make that decision to make the final call.

At the very least, it should be left to a Scheme to enforce this at the Switch level, in which case it should be made configurable (in the implementation) to enable a Scheme to choose whether to enforce it or not (at quotes and/or at transfers level)

This comes out of this excerpt from the API Definition Section **6.7.1.4**, especially the **note**

```
The Payer FSP must always set a transfer expiry time to allow for use cases in which a swift completion or failure is needed. If the use case does not require a swift completion, a longer expiry time can be set. If the Payee FSP fails to respond before the expiry time, the transaction is cancelled in the Payer FSP. The Payer FSP should still expect a callback from the Payee FSP.
Short expiry times are often required in retail scenarios, in which a customer may be standing in front of a merchant; both parties need to know if the transaction was successful before the goods or services are given to the customer.

In Figure 51, an expiry has been set to 30 seconds from the current time in the request from the Payer FSP, and to 20 seconds from the same time in the request from the Switch to the Payee FSP. This strategy of using shorter timeouts for each entity in the chain from Payer FSP to Payee FSP should always be used to allow for extra communication time.

Note: It is possible that a successful callback might be received in the Payer FSP after the expiry time; for example, due to congestion in the network. The Payer FSP should allow for some extra time after the actual expiry time before cancelling the financial transaction in the system. If a successful callback is received after the financial transaction has been cancelled, the transaction should be marked for reconciliation and handled separately in a reconciliation process.
```

bushjames:
hi @elnyry, this behaviour is being included in the scheme-adapter so the DFSP can choose whether to reject requests based on expiry configuration options. Enforcement (or not) at the switch is not relevant to this particular change. This just allows an SDK using switch to say "I want to reject responses that dont meet my deadlines" etc... hope that clarifies.

elnyry:
ok thanks @bushjames .. DFSP side is good

src/outboundApi/handlers.js
```diff
@@ -40,7 +40,7 @@ const postTransfers = async (ctx) => {
         ctx.response.body = response;
     }
     catch(err) {
-        ctx.response.status = 500;
+        ctx.response.status = err.httpStatusCode || 500;
```
bushjames: I think you also need to add this change to the putTransfers handler - if we are in 3 or 2 stage mode, model might throw on missed deadlines also?

PR #65, opened 2019-10-16
Converted OpenAPI spec file to unix format

kirgene:


bushjames:
Closing so commits can be cherry picked into PR#64.

elnyry:
@bushjames 504 isn't defined in the possible HTTP codes in v1.0 of the Spec (for the [ML FSPIOP API](https://github.com/mojaloop/mojaloop-specification/blob/master/API%20Definition%20v1.0.pdf), Section 3.2.4); in case it comes under that one. If not, of course that restriction is not there.

bushjames:
@elnyry this is the "simplified" backend API on the scheme adapter, not the ML interface.

PR #72, opened 2019-11-08
MP-643 apply circular ref fixes to accounts model

bushjames:


src/lib/model/AccountsModel.js
```diff
@@ -352,21 +356,33 @@ class AccountsModel {
...
             this.logger.log(`Error running account model: ${util.inspect(err)}`);
-            await this._unsubscribeAll();
-            await this.stateMachine.error(JSON.parse(JSON.stringify(err)));
-            err.executionState = this.getResponse();
+
+            // as this function is recursive, we dont want to error the state machine multiple times
+            if(this.data.currentState !== 'errored') {
+                await this._unsubscribeAll();
+
+                // err should not have a transferState property here!
```
kirgene: should be `executionState` I think

PR #77, opened 2019-11-12
Replace toLowercase() in config.js which fails if variable is not defined in the env

ramirogm:


ramirogm:
Do we have a set of required vars? I can add that logic.
Currently `config.js` provides a set of DEFAULT values so this patch is targeted to not fail if the env var is not provided, as it will take the default

bushjames:
Closing as new config loading method is in use which should give desired behaviour.

PR #82, opened 2019-11-19
add data path to validation error messages to aid debugging

bushjames:


src/test/unit/OutboundServer.test.js
```diff
@@ -0,0 +1,105 @@
...
+    is(...args) {
+        console.log(`DummyRequest.is called with args: ${util.inspect(args)}`);
+
+        // simulate json
+        if(Array.isArray(args[0]) && args[0][0] === 'application/json') {
+            return true;
+        }
+        return false;
+    }
+}
```
kirgene: It seems to be the same class as from the `src/test/unit/InboundServer.test.js`, so it can be extracted into separate file.

src/test/unit/InboundServer.test.js
```diff
@@ -11,7 +11,8 @@
 'use strict';
 
 const util = require('util');
-const Readable = require('stream').Readable;
+const DummyRequest = require('./dummyRequest.js');
```
kirgene: I think `.js` suffix is redundant here.
Also, I think that it will be better to use Pascal case for filenames containing types (`DummyRequest.js`).

bushjames: facepalm. sigh. if you insist.

PR #83, opened 2019-11-19
Run linting in circleci on commit/merge

bushjames:


src/test/unit/OutboundServer.test.js
```diff
@@ -0,0 +1,105 @@
...
+    is(...args) {
+        console.log(`DummyRequest.is called with args: ${util.inspect(args)}`);
+
+        // simulate json
+        if(Array.isArray(args[0]) && args[0][0] === 'application/json') {
+            return true;
+        }
+        return false;
+    }
+}
```
kirgene: It seems to be the same class as from the `src/test/unit/InboundServer.test.js`, so it can be extracted into separate file.

PR #85, opened 2019-11-22
MP-728,MOWDEV-3218 use transferAmount from quote response in prepare

bushjames:
MP-728,MOWDEV-3218 use transferAmount from quote response in prepare, not amount from initial request

### DIFF
--- a/src/test/unit/lib/model/OutboundTransfersModel.test.js
+++ b/src/test/unit/lib/model/OutboundTransfersModel.test.js
@@ -157,6 +157,7 @@ describe('outboundModel', () => {
             cache.emitMessage(JSON.stringify(payeeParty));
             return Promise.resolve();
         });
+
         MojaloopRequests.__postQuotes = jest.fn(() => {
             // ensure that the `MojaloopRequests.postQuotes` method has been called with correct arguments
             // including extension list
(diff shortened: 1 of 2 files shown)


src/test/unit/lib/model/OutboundTransfersModel.test.js
```diff
@@ -215,6 +216,98 @@ describe('outboundModel', () => {
...
+
+        const dummyTransferAmount = {
+            currency: 'XYZ',
+            amount: '9876543210'
+        };
+
+        MojaloopRequests.__postQuotes = jest.fn(() => {
+            // ensure that the `MojaloopRequests.postQuotes` method has been called with correct arguments
+            // including extension list
+            expect(MojaloopRequests.__postQuotes).toHaveBeenCalledTimes(1);
```
kirgene: I think this should placed outside of the function body, after `await model.run();`

src/test/unit/lib/model/OutboundTransfersModel.test.js
```diff
@@ -215,6 +216,98 @@ describe('outboundModel', () => {
...
+            expect(extensionList.extension.length).toBe(2);
+            expect(extensionList.extension[0]).toEqual({ key: 'qkey1', value: 'qvalue1' });
+            expect(extensionList.extension[1]).toEqual({ key: 'qkey2', value: 'qvalue2' });
+
+            // simulate a callback with the quote response
+            // note that we will change the currency and amount of the quote response to ensure
+            // the transfer prepare uses those rather than the originally requested currency.
+
+            // first make sure nobody has changed the quote response data
+            expect(quoteResponse.data.transferAmount).toEqual({
```
kirgene: I suppose you wanted to check the `postQuotes` request arg
```MojaloopRequests.__postQuotes = jest.fn(request => { ... })```
as `quoteResponse` (defined as `const quoteResponse = require('./data/quoteResponse');`) is never changed.

Comment (bushjames) [reply src/test/unit/lib/model/OutboundTransfersModel.test.js]: I just want to make sure the value in the mock data has not been changed so the test remains valid. if someone changed the mock data to match the overridden amount, the test would no longer be valid but would pass.

File: src/test/unit/lib/model/OutboundTransfersModel.test.js [hunk 3]
```diff
@@ -209,6 +209,99 @@ describe('outboundModel', () => {
...
+            expect(extensionList.extension).toBeTruthy();
+            expect(extensionList.extension.length).toBe(2);
+            expect(extensionList.extension[0]).toEqual({ key: 'tkey1', value: 'tvalue1' });
+            expect(extensionList.extension[1]).toEqual({ key: 'tkey2', value: 'tvalue2' });
+
+            expect(MojaloopRequests.__postTransfers.mock.calls[0][1]).toBe(quoteResponse.headers['fspiop-source']);
+            expect(model.data.to.fspId).toBe(payeeParty.party.partyIdInfo.fspId);
+            expect(quoteResponse.headers['fspiop-source']).not.toBe(model.data.to.fspId);
+
+            expect(postTransfersBody.amount).toEqual(quoteResponse.data.transferAmount);
```
kirgene: I've just realized that this code block should also be placed after `await model.run();`, specifically after `.toHaveBeenCalledTimes` call.

bushjames: I dont think that can be done exactly @kirgene , these expects are checking the value passed to postTransfers.

kirgene: Yeah, these are the checks for `postTransfers` args, so I think it would be better to write something like this:
```
       MojaloopRequests.__postTransfers = jest.fn((body, destFspId) => {

...
            const extensionList = body.extensionList;
...
            expect(destFspId).toBe(quoteResponse.headers['fspiop-source']);
...
   });
 ```

PR #90, opened 2019-11-26
Feature/1068 transfer without quotes

vijayg10:
* Modified incoming transfers so that previous quote is not mandatory. This is to support transfer only requests for performance testing where we want to isolate the transfers only without parties and quotes.


vijayg10:
https://github.com/mojaloop/mojaloop-simulator/pull/19

vijayg10:
https://github.com/mojaloop/mojaloop-simulator/pull/19

PR #101, opened 2019-12-04
Updates for perf testing

mdebarros:
- Updated event-sdk dependency to support tracestate w3c trace header
- Added repository info to the package.json
- Merged latest changes from master branch


src/lib/model/OutboundTransfersModel.js
```diff
@@ -317,6 +317,9 @@ class OutboundTransfersModel {
             }, ASYNC_TIMEOUT_MILLS);
 
             // create a quote request
+            const quote = this._buildQuoteRequest();
+            this.data.quoteId = quote.quoteId;
+
```
rmothilal: this quote is created 12 lines above this, any reason why it is duplicated and moved here?

PR #115, opened 2019-12-23
Add API test cases, refactoring.

kirgene:
* Use `env-var` package for reading parsing config.
* Prefix private class fields with underscore.
* Move JWS keys watcher from config to `InboundServer`.
* Add `disconnect` method to `Cache`.
* Add test cases for outbound API.
* Refactoring.


src/config.js
```diff
@@ -10,218 +10,96 @@
 'use strict';
 
 const fs = require('fs');
-const path = require('path');
+require('dotenv').config({ path: 'local.env' });
```
bushjames: what does dotenv give us that we dont have already with docker (docker supports env files)?

src/OutboundServer/api.yaml
```diff
@@ -562,9 +562,7 @@ components:
           description: >
             Object representing the last error to occur during a transfer process. This may be a Mojaloop API error returned from another entity in the scheme
             or an object representing other types of error e.g. exceptions that may occur inside the scheme adapter.
-          oneOf:
-            - $ref: '#/components/schemas/transferError'
-            - $ref: '#/components/schemas/generalError'
+          $ref: '#/components/schemas/transferError'
```
bushjames: why did you make this change? I believe it is still possible for either of these to be returned?

src/lib/cache/cache.js
```diff
@@ -52,11 +52,26 @@ class Cache {
      * See: https://redis.io/topics/pubsub
      */
     async connect() {
-        this.client = await this._getClient();
-        this.subscriptionClient = await this._getClient();
+        if (this._connected) {
```
bushjames: This could be an issue if the caller deliberately want to reconnect. He will not know nothing is happening when he calls this func. Why not throw an exception instead?

src/config.js
kirgene: It's convenient during development when you run app directly on local machine without docker.

src/OutboundServer/api.yaml
kirgene: I haven't seen the code that could generate this general error

src/InboundServer/index.js
```diff
@@ -142,6 +143,48 @@ class InboundServer {
...
+
+    _startJwsWatcher() {
+        const FS_EVENT_TYPES = {
+            CHANGE: 'change',
+            RENAME: 'rename'
+        };
+        const watchHandler = (eventType, filename) => {
+            // On most platforms, 'rename' is emitted whenever a filename appears or disappears in the directory.
+            // From: https://nodejs.org/docs/latest/api/fs.html#fs_fs_watch_filename_options_listener
+            if (eventType === FS_EVENT_TYPES.RENAME) {
```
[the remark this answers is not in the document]
kirgene: fixed

src/InboundServer/index.js
```diff
@@ -142,6 +143,48 @@ class InboundServer {
...
+                const keyName = path.basename(filename, '.pem');
+                if (this._jwsVerificationKeys[keyName] == null) {
```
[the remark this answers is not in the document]
kirgene: fixed

elnyry:
@kirgene can you please update the version to v8.8.0.

Here are the guidelines: https://github.com/mojaloop/documentation/blob/master/contributors-guide/standards/versioning.md

src/lib/cache/cache.js
kirgene: done

PR #122, opened 2020-01-10
Fix/1134 redis cache expiry

lewisdaly:
- Adds an expiry option to the redis cache
- This will allow us to use the mojaloop-simulator without continually using disk space
- Set up integration tests for this repo


lewisdaly:
Hi @bushjames,thanks for your comments - I certainly underestimated that this was more of a database than a cache.

I think adding the memory limit will likely fix the issue you predicted we were having with disk usage, and not post such a risk for production use of the library, as we can specify the config within our test helm charts.

I'll close this PR for now, and make a new one for the docker-compose changes with some integration tests.

PR #125, opened 2020-01-13
Feature/add integration tests

lewisdaly:
- adds docker-compose example integration test config
- updates `Dockerfile` to use staged builds


bushjames:
@kirgene I quite like the approach of using docker compose to test with some real implementations; redis in this case but could be good for running one instance against another for example. Do you have any thoughts on this?

Dockerfile
```diff
@@ -8,7 +8,16 @@ COPY ./secrets /
 
 WORKDIR /src/
 
-CMD ["node", "/src/index.js"]
-
 COPY ./src/ /src/
-RUN npm install --production
+RUN npm install
+RUN npm rebuild
```
kirgene: I think the `rebuild` here is redundant as this docker layer will be removed as soon as `src` content changes.
Also, I think that we should install/build all packages only once in `builder` via `npm install --production` and then just copy the `src` dir to the final image.

Dockerfile
```diff
@@ -8,7 +8,16 @@ COPY ./secrets /
...
+
+FROM node:12.14.0-alpine 
+
+WORKDIR /src/
+
+COPY --from=builder /src/ .
+COPY ./secrets /
```
kirgene: Probably better to mount that directory at runtime. I don't think it's a good idea to copy anything named `secrets` or `keys` into a final docker image.
If this directory is to be used only for testing, then it's better to move it to `tests` dir.

Dockerfile
```diff
@@ -8,7 +8,16 @@ COPY ./secrets /
...
+RUN npm prune --production
```
kirgene: Please see the above comment.

Dockerfile
```diff
@@ -8,7 +8,16 @@ COPY ./secrets /
...
-CMD ["node", "/src/index.js"]
-
 COPY ./src/ /src/
-RUN npm install --production
+RUN npm install
+RUN npm rebuild
+
+FROM node:12.14.0-alpine 
+
+WORKDIR /src/
```
kirgene: I think it worth changing directory name to `/app`

Dockerfile
```diff
@@ -8,7 +8,16 @@ COPY ./secrets /
...
+
+COPY --from=builder /src/ .
+COPY ./secrets /
+RUN npm prune --production
+
+CMD ["node", "/src/index.js"]
```
kirgene: It can simply `CMD ["node", "/app"]`

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+  scheme-adapter:
+    build:
+      context: .
+    env_file: ./src/test/config/integration.env
+    ports:
+      - "4000:4000"
+      - "4001:4001"
+    depends_on:
+      - redis 
+
```
kirgene: please remove redundant new lines

docker-compose.integration.yml
```diff
@@ -0,0 +1,28 @@
...
+      - ./src:/src
+      - node_modules:/src/node_modules
+    depends_on:
+      - redis 
+    command:
+      - 'tail -f /dev/null'
+    entrypoint: 
+      - sh 
+      - -c
+
```
kirgene: please remove redundant new lines

Dockerfile
lewisdaly: Yeah we can probably remove the rebuild, I was keeping it around since it was already there.

As for the `npm install` step here, reason we do this plus the `npm prune --production` in the next step is so that we can have `devDependencies` still available for things such as integration tests during the builder stage. This just makes life a whole lot easier as then we only need 2 targets in 1 Dockerfile. Happy to hear other alternatives though.

Dockerfile
lewisdaly: Answered above

Dockerfile
lewisdaly: Yeah, I'm not 100% sure about what this is doing here and why it can't be mounted at runtime... @bushjames might have some better answers for us, but I suspect it simply makes life easier in non-production environments.

Dockerfile
kirgene: I see, so probably we can have all these steps inside `builder`, like
`npm install`
`npm test`
`npm prune --production`

Dockerfile
lewisdaly: Maybe... I'll let @bushjames chime in as I think there are a few repos standardized on `/src/`

Dockerfile
bushjames: @kirgene @lewisdaly the secrets directory gets copied to provide some "out of the box defaults". These are test secrets only, available on github etc... in production DFSPs must mount over this directory with their own secrets. These defaults facilitate testing without having to go through the pain of generating them.

Dockerfile
bushjames: @kirgene what is the rationale for changing /src to /app?

Dockerfile
lewisdaly: No, if you look at the `docker-compose.integration.yml` file, we build a docker image that targets the `builder` only, which has the necessary `devDependencies` to run the integration tests.

In the past we maintained 2 separate Dockerfiles, which was a real pain, both from docker caching, and also because our tests should be as close as possible to what gets put into the production docker containers.

Dockerfile
kirgene: @bushjames I think it's considered as best practice (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

Dockerfile
kirgene: makes sense, so probably just remove the `rebuild`

Dockerfile
lewisdaly: I don't think this really matters at the moment, plus it has other implications for the directory structure on other projects.

I propse we should leave it as is for now, and if you want @kirgene maybe we can take it up as a question of standardization across mojaloop repos containing Dockerfiles.

kirgene: Yes, it's actually just a suggestion, so we can leave it as-is for now.

kirgene:
@lewisdaly could you please remove `package-lock.json`, and `npm rebuild`

lewisdaly:
> @lewisdaly could you please remove `package-lock.json`, and `npm rebuild`

I'll get rid of the `npm rebuild` step -- but I'm pretty sure we don't want to get rid of the `package-lock.json` file...

bushjames:
@lewisdaly @kirgene I am ambivalent about package-lock.json being checked in. There is a big danger that devs commit their local copy which causes frequent issues when doing docker builds in a different architecture to the dev machine. I am pretty much of the opinion that these issues occur so frequently that it is not worth it. Also, with docker being "immutable" and "tested" as an image itself, I do not think it is necessary to check in package-lock.json.

kirgene:
> @lewisdaly @kirgene I am ambivalent about package-lock.json being checked in. There is a big danger that devs commit their local copy which causes frequent issues when doing docker builds in a different architecture to the dev machine. I am pretty much of the opinion that these issues occur so frequently that it is not worth it. Also, with docker being "immutable" and "tested" as an image itself, I do not think it is necessary to check in package-lock.json.

That's exactly what I often observe, `package-lock.json` creates more problems than solves.

lewisdaly:
> > @lewisdaly @kirgene I am ambivalent about package-lock.json being checked in. There is a big danger that devs commit their local copy which causes frequent issues when doing docker builds in a different architecture to the dev machine. I am pretty much of the opinion that these issues occur so frequently that it is not worth it. Also, with docker being "immutable" and "tested" as an image itself, I do not think it is necessary to check in package-lock.json.
> 
> That's exactly what I often observe, `package-lock.json` creates more problems than solves.

No worries, I just wanted to check we're on the same page -- I've seen good arguments both for and against checking it in.

PR #132, opened 2020-03-04
Feature/mp 940 transaction request

ksatya77:
Adding support for transactionRequest endpoint which is needed for Payee Initiated Transactions


src/test/unit/lib/model/OutboundRequestToPayModel.test.js
```diff
@@ -0,0 +1,754 @@
...
+        expect(MojaloopRequests.__postTransactionRequests).toHaveBeenCalledTimes(1);
+        
+        // check we stopped at payeeResolved state
+        expect(result.currentState).toBe('COMPLETED');
+        expect(result.transactionRequestResponse.transactionRequestState).toBe('RECEIVED');
+        expect(StateMachine.__instance.state).toBe('succeeded');
+    });
+
+
+    // test('uses quote response transfer amount for transfer prepare', async () => {
```
kirgene: I think better to remove such code in comments from this repository and leave it in only in your repo (in a different branch) if necessary.
For this repository, you can put some TODOs.

src/InboundServer/handlers.js
```diff
@@ -307,6 +352,29 @@ const putQuoteById = async (ctx) => {
...
+        const req = {
+            headers: ctx.request.headers,
+            data: ctx.request.body
+        };
+        const res = await ctx.state.cache.set(`callback_${ctx.state.path.params.ID}`, req);
+        ctx.state.logger.log(`Cacheing callback: ${util.inspect(res)}`);
+    }
+
+    // publish an event onto the cache for subscribers to action
+    await ctx.state.cache.publish(`tr_${ctx.state.path.params.ID}`, {
```
bushjames: @ksatya77 please use a different channel prefix for transaction requests. "tr_..." is already used by the outbound transfers model; even if there is only small probability of ID clash with transfers I would rather see something different in this model e.g. "trq_...".

src/lib/model/InboundTransfersModel.js
```diff
@@ -244,6 +275,45 @@ class InboundTransfersModel {
         }
     }
 
+    /**
+     * Validates  an incoming transfer prepare request and makes a callback to the originator with
+     * the result
+     */
+    async executeTransactionRequest(transactionRequest, sourceFspId) {
```
bushjames: @ksatya77 What is the puspose of executeTransactionRequest function? This looks like duplicate of transactionRequest functino in this file?

src/lib/model/OutboundRequestToPayModel.js
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    async _executeTransactionRequest() {
+        // eslint-disable-next-line no-async-promise-executor
+        return new Promise(async (resolve, reject) => {
+            // create a transfer prepare request
+            const transactionRequest = this._buildTransactionRequest();
+
+            // listen for events on the transactionRequestId
+            const transactionRequestKey = `tr_${this.data.transactionRequestId}`;
```
bushjames: @ksatya77 see my previous comment on cache channel name. Best not to reuse "tr_..." from the outbound transfers model due to (admittedly low) possibility of ID clash.

src/lib/model/InboundTransfersModel.js
ksatya77: ohh yes...copy paste issue...removed the duplicate now

src/lib/model/OutboundRequestToPayModel.js
ksatya77: renamed it to txnreq_

src/OutboundServer/api.yaml
```diff
@@ -121,6 +121,36 @@ paths:
...
+          2. Transaction Request. This request enables a Payee to request Payer to send electronic funds to the Payee. 
+          
+      tags:
+        - RequestToPay
+      requestBody:
+        description: RequestToPay request body
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/requestToPayInboundRequest'
```
bushjames: @ksatya77 isnt this a request to pay outbound request? not inbound is it?

bushjames: are you using exact same body for outgoing as incoming requests?

src/OutboundServer/api.yaml
```diff
@@ -581,6 +637,73 @@ components:
...
+        amount:
+          $ref: '#/components/schemas/money'
+        scenario:
+          $ref: '#/components/schemas/scenario'
+        initiator:
+          $ref: '#/components/schemas/initiator'
+        initiatorType:
+          $ref: '#/components/schemas/initiatorType'
+
+    requestToPayResponse:
```
bushjames: section 6.4.3.1 of API spec document suggests the response for a transactionRequest will contain a "transactionRequestState" object. I think you probably want to model and expose that to the DFSP backend in some way, rather than simply reflecting the request back to them.

ksatya77: yeah..good point...the payee should definitely know the status of this request so that he can proceed to next steps if it was accepted.

ksatya77: @bushjames ... I just checked the response being returned to Backend. The final state is indeed returned back.

{
    "transactionRequestId": "cd04d5fc-8328-4ca6-9761-fce75b1a3e2d",
    "from": {
        "idType": "BUSINESS",
        "idValue": "Starbucks",
        "displayName": "Starbucks Coffee"
    },
    "to": {
        "idType": "MSISDN",
        "idValue": "17039811907",
        "fspId": "payerfsp",
        "firstName": "GoldenFirst",
        "middleName": "GoldenMiddle",
        "lastName": "GoldenLast",
        "dateOfBirth": "2010-10-10"
    },
    "amountType": "SEND",
    "currency": "USD",
    "amount": "10",
    "scenario": "PAYMENT",
    "initiator": "PAYEE",
    "initiatorType": "BUSINESS",
    "note": "test payment",
    "currentState": "COMPLETED",
    "transactionRequestResponse": {
        "transactionRequestState": "RECEIVED"
    }
}

ksatya77: @bushjames are you suggesting "transactionRequestState": "RECEIVED" to be at the root level instead of within transactionRequestResponse?

ksatya77: @bushjames added a new field to represent the final state, requestToPayState

src/OutboundServer/api.yaml
```diff
@@ -581,6 +637,76 @@ components:
         transferRequestExtensions:
           $ref: '#/components/schemas/extensionList'
 
+    requestToPayInboundRequest:
```
bushjames: @ksatya77 Is this incoming to the DFSP or outgoing from the DFSP to switch?
If it is outgoing request sent by the DFSP using the scheme adapter I think the request ID should be generated in the SDK in same way as transferId is generated for outgoing transfers....the response to the backend should contain the ID so the DFSP backend can record it for later use if necessary.

src/OutboundServer/handlers.js
```diff
@@ -137,6 +140,33 @@ const postAccounts = async (ctx) => {
     }
 };
 
+const postRequestToPay = async (ctx) => {
+    try {
+        // this requires a multi-stage sequence with the switch.
+        let requestToPayInboundRequest = {
```
bushjames: @ksatya77 surely this is an outbound request? ... in context of scheme adapter:
- OUTBOUND means requests from the DFSP owning the scheme adapter instance to other participants (via switch)
- INBOUND means requests coming to the scheme adapter from other participants (via switch)

src/lib/model/OutboundRequestToPayModel.js
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /**
+     * Builds a transaction request payload from current state
+     *
+     * @returns {object} - the transaction request payload
+     */
+    _buildTransactionRequest() {
+        let transactionRequest = {
+            transactionRequestId: this.data.transactionRequestId,
```
bushjames: @ksatya77 see my other comment on generating the transactionRequestId here rather than requiring DFSP backend to generate it. Please do same as we do with transferId in outbound transfers model.

src/OutboundServer/handlers.js
```diff
@@ -156,4 +186,7 @@ module.exports = {
     '/accounts': {
         post: postAccounts
     },
+    '/requestToPay': {
+        post: postRequestToPay
```
bushjames: @ksatya77 I think you have forgotten to include a PUT handler for the outbound transfer request model. If the model halts after resolving payee there is no way for the DFSP backend to tell the scheme adapter to proceed with the next stage.

src/OutboundServer/api.yaml
ksatya77: Updated the schema names to be requestToPayRequest and requestToPayResponse to avoid confusion and also to align with the naming convention used for /transfers.

src/lib/model/OutboundRequestToPayModel.js
ksatya77: I now removed this as a required field in the incoming request.
// add a transactionRequestId if one is not present e.g. on first submission
        if(!this.data.hasOwnProperty('transactionRequestId')) {
            this.data.transactionRequestId = uuid();
        }
this adds the transactionRequestId and it is sent back in the response.

src/OutboundServer/handlers.js
ksatya77: I am planning to add this logic as part of a separate PR

src/test/unit/lib/model/data/requestToPayRequest.json
```diff
@@ -0,0 +1,26 @@
+{
```
bushjames: @ksatya77 are you sure this is right? It looks like an outbound transfer request rather than an outbound request to pay request.

ksatya77: Updated the request now...

PR #133, opened 2020-03-10
MP-637/MP-495: Implement GET /transfers API on inbound and outbound interfaces

kirgene:


src/lib/model/InboundTransfersModel.js
```diff
@@ -113,6 +117,38 @@ class InboundTransfersModel {
         }
     }
 
+    /**
+     * Queries details of a transfer
+     */
+    async getTransfer(transferId, sourceFspId) {
```
bushjames: @kirgene Incoming GET /transfer requests should not be sourced from the cache but should call out to the DFSP backend with a GET request to the transfers endpoint. (same endpoint as the POST incoming transfer endpoint the DFSP has to implement). It is possible that the DFSP backend will have changed the state of the transfer outside the cache lifetime, or that a request will simply be for a transfer not currently in the cache. It is not safe to assume the cache will have every transfer the DFSP participated in since the beginning of time.

The response from the backend should include all required elements to respond to the request. The call to the DFSP backend does not need to match mojaloop API data models but should follow the existing backend API data model. The response to the switch must be in mojaloop API model.

src/lib/model/InboundTransfersModel.js
```diff
@@ -207,13 +249,26 @@ class InboundTransfersModel {
...
             }
 
             // project the incoming transfer prepare into an internal transfer request
             const internalForm = shared.mojaloopPrepareToInternalTransfer(prepareRequest, quote);
 
+
+            await this._cache.set(`tf_${prepareRequest.transferId}`, {
+                request: prepareRequest,
+                mojaloopResponse: {
+                    transferState: 'RESERVED',
```
bushjames: @kirgene the DFSP backend must be the single source of truth for transfer state; the cache in the SDK should not really be thought of as that point of truth.

src/lib/model/InboundTransfersModel.js
```diff
@@ -228,9 +283,23 @@ class InboundTransfersModel {
             const mojaloopResponse = {
                 completedTimestamp: new Date(),
                 transferState: 'COMMITTED',
```
bushjames: @kirgene note that this is where the scheme adapter sets the transfer state as committed; the SDK should have no awareness of other states right now. If the call to the DFSP backend fails the scheme adapter returns error, if it succeeds it returns transfer as COMMITTED; there is no middle ground.

src/lib/model/OutboundTransfersModel.js
```diff
@@ -520,6 +524,103 @@ class OutboundTransfersModel {
         });
     }
 
+    /**
+     * Get transfer details by sending GET /transfers request to the switch;
+     * then waits for a notification from the cache that the transfer has been resolved.
+     */
+    async _getTransfer() {
```
bushjames: @kirgene I think you have misunderstood this story a bit. GET /transfer call from DFSP backend to switch should pass through the scheme-adapter, sure you need to turn async to sync but the transfer state must not come from the cache! The single source of truth for the transfer is the switch. This method must ALWAYS call the switch; your lines 537-548 are superfluous; there may NEVER be state in the cache for the requested transfer, that is a normal use case, scheme-adapter should ALWAYS forward the request to the switch even if it has never seen that ID before.

src/lib/model/lib/requests/backendRequests.js
```diff
@@ -44,7 +44,17 @@ class BackendRequests {
     async getParties(idType, idValue, idSubValue) {
         const url = `parties/${idType}/${idValue}`
           + (idSubValue ? `/${idSubValue}` : '');
-        return this._get(url, 'parties');
```
bushjames: @kirgene why this change?

kirgene: `_get` function has only one arg

src/lib/model/OutboundTransfersModel.js
```diff
@@ -520,6 +524,83 @@ class OutboundTransfersModel {
...
+                        this._logger.log(`Error unsubscribing (in callback) ${transferKey} ${subId}: ${e.stack || util.inspect(e)}`);
+                    });
+
+                    if (error) {
+                        return reject(error);
+                    }
+
+                    const fulfil = message.data;
+                    this._logger.push({ fulfil }).log('Transfer fulfil received');
+                    this.data.fulfil = fulfil;
```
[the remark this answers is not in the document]
kirgene: @bushjames these properties are in `fulfil` object. There is a [test](https://github.com/mojaloop/sdk-scheme-adapter/blob/414198fc53b42fb5d38a3e77541f1d9a7a16ff5f/src/test/unit/api/transfers/transfers.test.js#L183) that covers this case using [these](https://github.com/mojaloop/sdk-scheme-adapter/blob/414198fc53b42fb5d38a3e77541f1d9a7a16ff5f/src/test/unit/api/transfers/data/getTransfersCommittedResponse.json) data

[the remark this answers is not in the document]
bushjames: understood. thanks.

src/test/unit/api/accounts/utils.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ *
+ * @param reqInbound
+ * @param reqOutbound
+ * @param apiSpecsOutbound
+ * @return {function(...[*]=)}
+ */
+const createPostAccountsTester = ({reqInbound, reqOutbound, apiSpecsOutbound}) =>
+    /**
```
bushjames: @kirgene I think I would prefer to see the { } style with an explicit "return"...does this even work without?

src/config.js
```diff
@@ -64,6 +64,9 @@ module.exports = {
     autoAcceptQuotes: env.get('AUTO_ACCEPT_QUOTES').default('true').asBool(),
     autoAcceptParty: env.get('AUTO_ACCEPT_PARTY').default('true').asBool(),
 
+    /* TODO:  high-risk transactions can require additional clearing check */
```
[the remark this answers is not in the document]
kirgene: That's from API doc (6.7.1.6 Optional Additional Clearing Check), one of the cases when this GET /transfers API can be useful.

src/lib/model/InboundTransfersModel.js
```diff
@@ -113,6 +117,60 @@ class InboundTransfersModel {
         }
     }
 
+    /**
+     * Queries details of a transfer
+     */
+    async getTransfer(transferId, sourceFspId) {
+        try {
+            // make a call to the backend to get transfer details
+            const response = await this._backendRequests.getTransfers(transferId);
```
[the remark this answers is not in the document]
kirgene: added missing swagger doc

src/config.js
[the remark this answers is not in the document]
bushjames: ok @kirgene . Leaving commented out for now is fine.

PR #136, opened 2020-03-23
MP-1261 error extensionList as source of error body statusCode property

bushjames:
MP-1261 add config to allow an extensionList item in a mojaloop error callback to be the source of statusCode in synchronous backend error response.

### DIFF
--- a/src/OutboundServer/handlers.js
+++ b/src/OutboundServer/handlers.js
@@ -35,18 +35,17 @@ const handleError = (method, err, ctx, stateField) => {
         // by default we set the statusCode property of the error body to be that of any model state lastError
         // property containing a mojaloop API error structure. This means the caller does not have to inspect
         // the structure of the response object in depth to ascertain an underlying mojaloop API error code.
-        ctx.response.body.statusCode = err[stateField].lastError.mojaloopError.errorInformation.errorCode;
+        const errorInformation = err[stateField].lastError.mojaloopError.errorInformation;
+        ctx.response.body.statusCode = errorInformation.errorCode;
 
         // if we have been configured to use an error extensionList item as status code, look for it and use
         // it if it is present...
         if(ctx.state.conf.outboundErrorStatusCodeExtensionKey
-            && err[stateField].lastError.mojaloopError.errorInformation.extensionList
-            && Array.isArray(err[stateField].lastError.mojaloopError.errorInformation.extensionList)) {
+            && Array.isArray(errorInformation.extensionList)) {
 
             // search the extensionList array for a key that matches what we have been configured to look for...
             // the first one will do - spec is a bit loose on duplicate keys...
-            const extensionItem = err[stateField].lastError.mojaloopError.errorInformation
-                .extensionList.find(e => {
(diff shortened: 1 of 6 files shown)


src/OutboundServer/handlers.js
```diff
@@ -32,7 +32,28 @@ const handleError = (method, err, ctx, stateField) => {
...
 
+        // by default we set the statusCode property of the error body to be that of any model state lastError
+        // property containing a mojaloop API error structure. This means the caller does not have to inspect
+        // the structure of the response object in depth to ascertain an underlying mojaloop API error code.
         ctx.response.body.statusCode = err[stateField].lastError.mojaloopError.errorInformation.errorCode;
+
+        // if we have been configured to use an error extensionList item as status code, look for it and use
+        // it if it is present...
+        if(ctx.state.conf.outboundErrorStatusCodeExtensionKey
+            && err[stateField].lastError.mojaloopError.errorInformation.extensionList
```
kirgene: I think this check is redundant, `isArray` is sufficient

src/OutboundServer/handlers.js
```diff
@@ -32,7 +32,28 @@ const handleError = (method, err, ctx, stateField) => {
...
+            && Array.isArray(err[stateField].lastError.mojaloopError.errorInformation.extensionList)) {
+
+            // search the extensionList array for a key that matches what we have been configured to look for...
+            // the first one will do - spec is a bit loose on duplicate keys...
+            const extensionItem = err[stateField].lastError.mojaloopError.errorInformation
```
kirgene: I see that `err[stateField].lastError.mojaloopError.errorInformation` is common in 38-48 lines, so it can be assigned to `errorInformation` variable for better readability

src/test/unit/config.test.js
```diff
@@ -35,6 +37,19 @@ describe('config', () => {
...
     });
 
+    it('correctly parses OUTBOUND_ERROR_STATUSCODE_EXTENSION_KEY when set', () => {
+        process.env.OUTBOUND_ERROR_STATUSCODE_EXTENSION_KEY = outErrorStatusKey;
+        const config = require('../../config');
+        expect(config.outboundErrorStatusCodeExtensionKey).toEqual(outErrorStatusKey);
+    });
+
+    it('correctly parses OUTBOUND_ERROR_STATUSCODE_EXTENSION_KEY when NOT set', () => {
+        delete process.env.OUTBOUND_ERROR_STATUSCODE_EXTENSION_KEY;
```
kirgene: `delete ...` not needed as `process.env` get cleared in `beforeEach` and `afterEach`

bushjames: I dont agree. the code in beforeEach/afterEach does not clear the environment. If a value was already in the environment for OUTBOUND_ERROR_STATUSCODE_EXTENSION_KEY then it would not be removed.

kirgene: you're right. I see that the env is been saving in `beforeEach` and restored in `afterEach`. So I suspect that test cases run in parallel and share the same `process.env`.

PR #137, opened 2020-03-27
MP-1140: Implemented Proxy for Reporting API.

kirgene:


src/OutboundServer/middlewares.js
```diff
@@ -74,8 +75,35 @@ const createRequestValidator = (validator) => async (ctx, next) => {
     }
 };
 
+/**
+ * Create proxy middleware for forwarding matching DFSP requests
+ * (from provided routing rules) to corresponding switch endpoint
+ * @param proxyConfig
+ * @return {Function}
+ */
+const createProxy = (proxyConfig) => async (ctx, next) => {
```
bushjames, line 84: @kirgene rather than instantiate a new proxy instance for every outgoing request why not create one proxy instance at higher level that handles all requests?

kirgene: @bushjames done

PR #144, opened 2020-04-13
Moved "test" API to a different server.

msk-:
A couple of other changes in this PR (although they probably shouldn't be, but I can't be bothered moving them):
- Dockerfile modified to cache node modules and avoid npm install every build
- Labels added to Dockerfile
- Added `.dockerignore` file to reduce docker build context size and avoid pollution by `node_modules`

Before merging, please review this:
https://github.com/msk-/sdk-scheme-adapter/pull/1

Additionally, it looks like the integration tests are not passing- was that the case previously? Is it because I changed the `npm install` command in the Dockerfile to `npm install --production`? I would have _guessed_ not, but I guess it's possible `npm prune --production` doesn't do what it says it does.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,15 +67,27 @@ defaults_build_docker_login: &defaults_build_docker_login
 
 defaults_build_docker_build: &defaults_build_docker_build
   name: Build Docker $CIRCLE_TAG image
-  command: |
+  command: >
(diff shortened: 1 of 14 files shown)


msk-:
James expressed some currently unspecified objection to the PR mentioned in the description so I haven't merged that at present.

PR #145, opened 2020-04-16
strip any base path off peerEndpoint config when proxying

bushjames:
strip any base path off peerEndpoint config setting when proxying requests

### DIFF
--- a/src/lib/model/ProxyModel/index.js
+++ b/src/lib/model/ProxyModel/index.js
@@ -31,6 +31,7 @@ class ProxyModel {
      */
     constructor(config) {
         this._logger = config.logger;
+
         this._requests = new MojaloopRequests({
             logger: this._logger,
             peerEndpoint: config.peerEndpoint,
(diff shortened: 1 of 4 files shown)


src/lib/model/ProxyModel/index.js
```diff
@@ -32,6 +32,10 @@ class ProxyModel {
     constructor(config) {
         this._logger = config.logger;
 
+        //Note that we strip off any path on peerEndpoint config after the origin.
```
kirgene, line 35: I think this should be moved to `src/OutboundServer/index.js` also.

src/lib/model/ProxyModel/index.js
```diff
@@ -31,7 +31,6 @@ class ProxyModel {
      */
     constructor(config) {
         this._logger = config.logger;
-
```
kirgene, line 34: probably you can just revert this file from master.

PR #154, opened 2020-05-05
Correctly connecting/disconnecting to/from cache

msk-:


src/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "9.4.11",
+  "version": "9.4.12",
```
sri-miriyala, line 3: Is this not 10.1.x?

elnyry: Thank you @sri-miriyala it should be :-) .. Here's the schedule for future reference: https://github.com/mojaloop/documentation/blob/master/contributors-guide/standards/versioning.md#sprint-schedule-for-pi0

msk-:
```suggestion
  "version": "10.1.0",
```

PR #158, opened 2020-05-14
Fix excessive log verbosity

kirgene:


msk-:
Does it make sense to log some of the response data? At least the response code, perhaps also the body/headers?

kirgene:
@msk- I've added logs for response status code and headers

PR #159, opened 2020-05-15
#1325: Bulk quotes and bulk transfers support

oderayi:
- Added support for bulk quotes and bulk transfers.


elnyry:
Looks good to me, thanks @oderayi .. @rmothilal thoughts?

src/lib/model/index.js
```diff
@@ -13,6 +13,8 @@
 
 const InboundTransfersModel = require('./InboundTransfersModel.js');
 const OutboundTransfersModel = require('./OutboundTransfersModel.js');
+const OutboundBulkQuotesModel = require('./OutboundBulkQuotesModel');
+const OutboundBulkTransfersModel = require('./OutboundBulkTransfersModel.js');
```
bushjames, line 17: I would favour an approach where outbound bulk transfers are modelled the same way as outbound single transfers; i.e. in a single state machine model. Is there a reason to separate them?

src/lib/model/lib/shared/index.js
```diff
@@ -231,7 +233,7 @@ const internalTransactionRequestResponseToMojaloop = (internal) => {
  */
 const mojaloopPrepareToInternalTransfer = (external, quote) => {
     let internal = null;
-    if(quote) {
+    if (quote) {
```
bushjames, line 236: These unnecessary whitespace changes make diffs incredibly large and hard to read.

src/lib/model/lib/shared/index.js
```diff
@@ -275,6 +277,171 @@ const mojaloopTransactionRequestToInternal = (external) => {
...
+        // create a map of internal individual quotes payees indexed by quotedId, for faster lookup
+        const internalQuotesPayeesByQuoteId = {};
+        
+        for (const quote of bulkQuotes.internalRequest.individualQuotes) {
+            internalQuotesPayeesByQuoteId[quote.quoteId] = quote.to;
+        }
+        
+        // create a map of external individual transfers indexed by quotedId, for faster lookup
+        const externalTransferIdsByQuoteId = {};
+        const ilp = new Ilp({ secret: ilpSecret });
```
bushjames, line 410: ilpSecret never changes during the lifetime of the scheme adapter process. It is more efficient to Instantiate the ilp utility class once only, not every time a function is called. It would be better to not pass the ilpSecret to this function, instead pass an instance of the ilp class managed at the next layer up, or better still, pass the `getTransactionObject` function from the upper layer instance itself.

src/lib/model/index.js
oderayi: @bushjames This is a temporary design pending availability of bulk account lookup. I had earlier implemented this as a single state machine model but it was later agreed that the implementation be broken up.

src/lib/model/InboundTransfersModel.js
```diff
@@ -370,6 +315,356 @@ class InboundTransfersModel {
...
+                if (this._checkIlp && (transfer.condition !== condition)) {
+                    const transferError = this._handleError(new Error(`ILP condition in bulk transfers prepare for ${transfer.transferId} does not match quote`));
+                    individualTransferErrors.push({ transferId: transfer.transferId, transferError });
+                }
+            }
+
+            if (bulkQuote && this._rejectTransfersOnExpiredQuotes) {
+                const now = new Date().toISOString();
+                const expiration = bulkQuote.mojaloopResponse.expiration;
+                if (now > expiration) {
```
bushjames, line 538: @oderayi are you sure this is a valid comparison? are both these vars Date objects?

oderayi: @bushjames The two variables were in the simplified ISO 8601 format and so were comparable. I have however converted both to Date objects before comparing being the recommended JavaScript style.

PR #160, opened 2020-05-29
Do not parse already processed JSON response.

kirgene:
Recently introduced request library returns already parsed JSON response by default.


kleyow:
:ship:

PR #161, opened 2020-06-01
 Updated library to adapt dfsp party accounts.

kleyow:
Slightly confused here 
docs/dfspInboundApi.yaml - Api for calls from dfsp backend?
src/InboundServer/api.yaml - Api for icalls from switch?
src/OutboundServer/api.yaml - Api for calls sent to switch?
@bushjames @lewisdaly

### DIFF
--- a/docs/dfspInboundApi.yaml
+++ b/docs/dfspInboundApi.yaml
@@ -380,6 +380,13 @@ components:
       minItems: 0
       maxItems: 16
 
+    accountList:
+      type: array
+      items:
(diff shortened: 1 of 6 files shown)


kleyow:
Ok. It's seems that the docs may be in a confusing state.
I'm going to leave this as it is since this is branched off for PISP. 
Don't want to hold this back, I can ticket a revisit of the documentation.

Sorry for the white space diff :sweat_smile:

PR #169, opened 2020-06-11
Pisp/feat/305 authorizations

eoln:
Work in progress.
Goal: implement POST /authorization on Outbound Service


src/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "9.4.11",
+  "version": "10.1.5",
```
lewisdaly, line 3: This should be `10.4.0`, as we are in sprint 4 atm

eoln:
this PR is dropped and continued: https://github.com/mojaloop/sdk-scheme-adapter/pull/170

PR #173, opened 2020-06-12
pre commit/push hooks

eoln:
- setup pre-push hook using husky:
  - run unit tests
- setup pre-commit hook:
  - lint only staged files
- move jest configuration from the command line to the external file
- bring package-lock.json back and use `npm ci` in CI/CD

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ jobs:
           name: Delete build dependencies
           command: apk del build-dependencies
       - save_cache:
-          key: dependency-cache-5-{{ checksum "src/package.json" }}
+          key: dependency-cache-{{ checksum "src/package.json" }}
           paths:
             - src/node_modules
 
@@ -93,7 +93,7 @@ jobs:
           command: *defaults_Dependencies
       - restore_cache:
           keys:
-          - dependency-cache-5-{{ checksum "src/package.json" }}
+          - dependency-cache-{{ checksum "src/package.json" }}
       - run:
           name: Execute unit tests
           command: cd src && npm run test
@@ -135,7 +135,7 @@ jobs:
           command: *defaults_Dependencies
       - restore_cache:
           keys:
-          - dependency-cache-5-{{ checksum "src/package.json" }}
+          - dependency-cache-{{ checksum "src/package.json" }}
       - run:
           name: Create dir for lint results
           command: cd src && mkdir -p /lintresults
@@ -155,7 +155,7 @@ jobs:
           <<: *defaults_license_scanner
       - checkout
       - restore_cache:
-          key: dependency-cache-5-{{ checksum "src/package.json" }}
+          key: dependency-cache-{{ checksum "src/package.json" }}
       - run:
           name: Run the license-scanner
           command: cd /tmp/license-scanner && pathToRepo=$CIRCLE_WORKING_DIRECTORY make run
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,8 +74,8 @@ jobs:
           name: Access npm folder as root
           command: cd $(npm root -g)/npm
       - run:
-          name: Update NPM install
-          command: cd src && npm install
+          name: NPM installs dependencies in CI mode
+          command: cd src && npm ci
       - run:
           name: Delete build dependencies
           command: apk del build-dependencies
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ jobs:
           name: Delete build dependencies
           command: apk del build-dependencies
       - save_cache:
-          key: dependency-cache-{{ checksum "src/package.json" }}
+          key: dependency-cache-{{ checksum "src/package-lock.json" }}
           paths:
             - src/node_modules
 
@@ -93,7 +93,7 @@ jobs:
           command: *defaults_Dependencies
       - restore_cache:
           keys:
-          - dependency-cache-{{ checksum "src/package.json" }}
+          - dependency-cache-{{ checksum "src/package-lock.json" }}
       - run:
           name: Execute unit tests
           command: cd src && npm run test
@@ -135,7 +135,7 @@ jobs:
           command: *defaults_Dependencies
       - restore_cache:
           keys:
-          - dependency-cache-{{ checksum "src/package.json" }}
+          - dependency-cache-{{ checksum "src/package-lock.json" }}
       - run:
           name: Create dir for lint results
           command: cd src && mkdir -p /lintresults
@@ -155,7 +155,7 @@ jobs:
           <<: *defaults_license_scanner
       - checkout
       - restore_cache:
-          key: dependency-cache-{{ checksum "src/package.json" }}
+          key: dependency-cache-{{ checksum "src/package-lock.json" }}
       - run:
           name: Run the license-scanner
           command: cd /tmp/license-scanner && pathToRepo=$CIRCLE_WORKING_DIRECTORY make run
--- a/src/.npmrc
+++ b/src/.npmrc
@@ -1,1 +1,0 @@
-package-lock=false
--- a/src/package.json
+++ b/src/package.json
@@ -6,10 +6,14 @@
   "scripts": {
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
-    "test": "jest --ci --reporters=default --reporters=jest-junit --env=node test/unit",
-    "test:int": "jest --ci --reporters=default --reporters=jest-junit --env=node test/integration --forceExit"
+    "test": "jest --ci test/unit",
+    "test:int": "jest --ci --collectCoverage=false test/integration --forceExit"
+
   },
   "author": "Matt Kingston, James Bush, ModusBox Inc.",
+  "contributors": [
+    "Paweł Marzec <pawel.marzec@modusbox.com>"
+  ],
   "license": "Apache-2.0",
   "licenses": [
     {
@@ -17,6 +21,15 @@
       "url": "http://www.apache.org/licenses/LICENSE-2.0"
     }
   ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged && npm run test",
+      "post-commit": "git update-index --again"
+    }
+  },
+  "lint-staged": {
+    "*.{js,ts}": "eslint --cache --fix"
+  },
   "dependencies": {
     "@internal/cache": "file:lib/cache",
     "@internal/log": "file:lib/log",
@@ -43,8 +56,10 @@
     "eslint": "^7.0.0",
     "eslint-config-airbnb-base": "^14.1.0",
     "eslint-plugin-import": "^2.20.2",
+    "husky": "^4.2.5",
     "jest": "^26.0.1",
     "jest-junit": "^10.0.0",
+    "lint-staged": "^10.2.9",
     "nock": "^12.0.3",
     "openapi-response-validator": "^4.0.0",
     "redis-mock": "^0.49.0",
--- a/src/package.json
+++ b/src/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "10.1.5",
+  "version": "10.1.6",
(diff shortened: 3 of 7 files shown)


.circleci/config.yml
```diff
@@ -135,7 +135,7 @@ jobs:
           command: *defaults_Dependencies
       - restore_cache:
           keys:
-          - dependency-cache-5-{{ checksum "src/package.json" }}
+          - dependency-cache-{{ checksum "src/package.json" }}
```
KamuelaFranco, line 138: Can you make this rely on `package-lock.json`?

.circleci/config.yml
```diff
@@ -155,7 +155,7 @@ jobs:
           <<: *defaults_license_scanner
       - checkout
       - restore_cache:
-          key: dependency-cache-5-{{ checksum "src/package.json" }}
+          key: dependency-cache-{{ checksum "src/package.json" }}
```
KamuelaFranco, line 158: Can you make this rely on `package-lock.json`?

src/package.json
```diff
@@ -1,22 +1,34 @@
...
+    "Paweł Marzec <pawel.marzec@modusbox.com>"
+  ],
   "license": "Apache-2.0",
   "licenses": [
     {
       "type": "Apache-2.0",
       "url": "http://www.apache.org/licenses/LICENSE-2.0"
     }
   ],
+  "husky": {
```
KamuelaFranco, line 23: Strongly protest the use of pre-commit hooks like this, as I think with a CICD pipeline in place, they incur more cost than they provide in benefit.

.circleci/config.yml
eoln: you are right package-lock.json is a much better source of checksum! fixed!

src/package.json
eoln: I would insist to automate the control of code quality as early in the development workflow as possible.

.circleci/config.yml
eoln: fixed

src/package.json
vbarzokas: I like the idea but I would prefer this to be a `pre-push` hook instead to speed up development cycle, as there can be scenarios that you want to commit locally many times before going through quality control of your code.

KamuelaFranco: > I would insist to automate the control of code quality as early in the development workflow as possible.

I think it's a bit too prescriptive of a workflow. E.g., if you could dictate we all used VS Code and then dictate with a file that we all did a test run and a lint on every save, you'd probably be in favor of that. I wouldn't though because there's a certain point that you have to step back and let individual developers determine how they'd best like to get to the same conclusion, which is quality code that passes peer-review.

lewisdaly: I somewhat agree with this - especially in the use of the `npm-check-updates` that we have in other repos (which isn't in scope here).

My only other thing to add is this: `git commit -n` skips the pre-commit hooks, so you can somewhat get the best of both worlds.

eoln: @vbarzokas I moved unit testing to pre-push hook and leave linting - but only staged files - in pre-commit hook.

KamuelaFranco: Again, the change to a pre-push hook was made to be prescriptive of one person's particular workflow at the expense of another's--it does not serve to solve anything universally.

src/.npmrc
```diff
@@ -1 +0,0 @@
-package-lock=false
```
kirgene, line 1: Why the need of bringing back `package-lock.json`?

eoln: @kirgene I have had faced today problems with failing CI/CD because there where the use of `npm install` in CI/CD step and the checksum of package.json has been changed (because npm install updates the dependencies versions in package.json) so in the next steps after `setup` there was not possible to download cached value of node_modules. It is a very useful file. It gives you the guarantee that dependencies tree is frozen - the only problem is when you try to resolve the commit conflicts in this file manually instead of regenerate it by running `npm install` - I put a more elaborative explanation about this on slack today.

eoln: @kirgene could you give me the explanation of why not to use package-lock.json?

kirgene: @eoln 
npm install should not update package.json file. This might be an error somewhere in CI/CD scripts.
And regarding package-lock.json: I have a lot of cases when devs have struggled with dependency issues in projects that use package-lock.json. Usually, the process took a few hours investigating why things behave differently on different machines. The problem was in package-lock.json that contained incorrect dependencies (comparing to package.json). So the solution was usually to get rid of package-lock.json.
package-lock.json is generated from npm install (based package.json). When you generate package-lock.json for the first time it will probably contain correct dependencies, but when you then update your dependencies in package.json, then rerun npm install, things will usually break.
What I usually observe is that each PR usually contains a version change of some library. In that case, a developer would remove package-lock.json and regenerate it from scratch losing all previous "working" dependencies.
So better to have only one single source of truth for the dependencies. package-lock.json introduces more cons than pros.

msk-: I was happy to see `package-lock.json` in this repo. As you know Yev, I accept your reservations about `package-lock.json`, but disagree with your conclusion here:
> more cons than pros
mostly for security reasons, and secondarily because of regressions due to uncontrolled package version changes.

So this comment is here to lend support for the presence of `package-lock.json`.

kirgene: @msk- 
I understand your concerns about security because of the potential change of packages version.
But let's look at this potential issue from a different perspective.
First, I don't think that we are talking about the change of package version to a potentially vulnerable one, because we can't audit all codebase of node_modules directory.
And if we can't audit it, then any package version change should be considered safe (or at least acceptable).
If by security we mean breaking things because of (minor) version change, then that's also shouldn't be a problem as we have test cases run phase during the building of docker image. Honestly, I haven't ever seen a case when a minor version change caused any breaking issue (assuming that there are no tightly coupled dependencies with internal implementation of other packages).
And the last thing, after building a docker image we have an immutable snapshot of a library with all its dependencies. So every user of it will use the exact dependency tree.
So taking into account the above statements I can't see any security issue we could face using the current approach with having the package.json as the only dependency definition.

eoln: @kirgene thank you for your response, let me share my opinion: 
`npm install` will update dependencies according to sem-ver strategy in package.json - that exactly happened to me today - so cache value of npm_modules was not accessible

the common problem is that devs don't understand well why package-lock.json feature has been introduced and how to work with it in the correct way. As you pointed out: the common problem when working with package-lock.json is that dev is changing package.json dependencies manually without recreating package-lock.json <- I have had seen it so many times... A good example of the pre-commit guard is to check file stats of package.json and package-lock.json and refuse to commit package-lock.json when it is older than package.json. 

I strongly disagree with your opinion about security reasons. Keep your dependency tree frozen preserve security and keeping it volatile doesn't. The antipattern of using `npm install` in CI/CD workflow would be the real evil by allowing the hidden and uncontrolled updates of dependencies in the publication pipeline when you want them to be strict. Some teams secure themself against by using strict versions for dependencies - but this is only a workaround from the time when there was no locking feature in npm.
6 min reading for better understanding: 
https://medium.com/@Cyclodex/demystifying-npm-install-npm-ci-package-lock-json-2807fc0ee404

kirgene: @eoln 
Thanks, let me also share additional observations regarding the usage of package-lock.json.

When devs working at some code (within a single commit) and during development decide to update package.json because of requirements (manually, with the help of IDE or other tools like npm-check-updates) they right away execute `npm install` and If their new code doesn't work, they start to fix it (at that moment the issues could come from different sides - from code or incorrect dependencies in node_modules after "npm install"). When a dev realizes (though not immediately) that the error could come from package-lock.json, he removes the file and runs `npm install` once again. But it won't help usually, because the dependencies in node_modules have been already broken. And when the person tries to check the execution of code once again, he gets the same error and decides that error is in the code and starts to fix it. During this phase, the dev could involve a repository owner to help him understand why he gets such kind of issue, and thus already 2 people work on the same issue. After spending some time and blaming the world the person decides to remove the node_modules directory or even the entire project and iteratively adds new changes, line by line, to check what exactly caused the breakage. When the person finishes all these manipulations a few hours could have passed for nothing. And these all happen obviously before the person decides to "commit" his changes.

I've observed these situations so many times, and considering how much time was spent on struggling with such kinds of issues the best decision was to get rid of this package-lock.json completely.
IMHO the idea of having a fixed version of packages is great, but the actual implementation is a total failure.

And regarding the security, I don't see the difference between changing packages version while doing "npm install" on devs machine and doing so on CI/CD machine. Yes, it would be great to commit the exact dependency tree, but if later a developer decides to change just a single package version, he has to remove package-lock.json and recreate it (with npm install) and thus all packages version will be changed (here is the actual security issue). So it could seem nice that you have fixed dependencies in a repo, but in fact, in actively developing projects this doesn't protect from failures and security problems, considering that project codebase evolves each day.

lewisdaly: Hi All. Thanks for the discussion points. I see pros and cons on either side here, but I would like to zoom out a little bit as I see similar debates coming up from time to time in the Mojaloop community.

One of the pros of microservices is that separate dev teams can work in isolation from one another, and only really need to worry about the interfaces that their services expose and consume from one another. 

Therein also lies the challenge for other dev teams that need to work _across_ services, like the PISP guys, or like the Mojaloop OSS devs who are supporting adopted components.

We've had similar ~arguments~ debates before about such things as HTTP libraries (hapi vs koa), indentation (4 or 2 spaces)... the list goes on. 

My view is this: If it's in the Mojaloop Open Source project, it needs to be maintainable _by_ the Mojaloop community. The more fragmentation we have between projects, in things like style, frameworks, coding conventions, etc. the more time we waste. In the ideal microservices world, we could have a high degree of separation between teams, and teams would be able to choose whatever libraries, styles, tooling, workflows etc. that they would like. 

However, we are not in the ideal microservices world, we are in a community where there must always be some give and take in order to get along and be productive together. 

I would also draw your attention to the Standards guide we have [here](https://docs.mojaloop.io/documentation/contributors-guide/standards/), which doesn't have a section about `package-lock.json` files. Indeed I really don't want to have to add a section about `package-lock.json` files because once we start adding sections and debating upon every little thing, we will have little time left for anything else.

msk-: Actually, it occurs to me that perhaps I haven't observed issues with `package-lock.json` for some time because I nearly never use `npm install` now. I almost always use `npm ci`.

Anyway, I'll concretely specify the main security concern I have. As the authors of a financial services platform there's a very real threat from the following scenario:
1. we use package `js-dependency: 1.0.0`
2. we publish our `package.json` in a publicly visible repository
3. ownership of package `js-dependency` is assumed by a bad actor
4. `js-dependency: 1.0.1` is published with malicious code
5. we execute `npm install` during our build process, and `js-dependency: 1.0.1` is installed

`package-lock.json` prevents this by constraining and checksumming installed packages.

This can also be taken as an argument for heavily auditing, vetting and (in aid of avoiding the effort of auditing and vetting) minimising dependencies.

eoln: Yes, Yes @msk. But there is a need to be aware that the lock-files (npm/yarn) can also be used to realized attack on the software supply chain, more here: https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/
to protect against this vector we should consider using the lock-file linting in our ci/cd pipelines by using this tool for example https://github.com/lirantal/lockfile-lint

msk-: Yes, it's definitely good to be aware of that. And would be better to incorporate it into CI/CD (in a separate build stage!). Tangentially, I notice this linked from that repo: https://github.com/lirantal/npq. I like that it gives you a basic sanity check, but I wonder if it also takes responsibility from the user when it shouldn't.

msk-:
I'd like to see progress on this PR @kirgene @bushjames. I think you should make a clear statement one way or the other about `npm ci` and `package-lock.json` so that the rest of this PR can move forward.

src/package.json
bushjames: I am also of the opinion that the branch protections and CI should take the responsibility of "enforcing" the quality of master branch. I think that although some may find git hooks useful, others will find them intrusive. This is a community though but the standards do not appear to specify one way or the other. I suggest we look for the best compromise.

PR #174, opened 2020-06-30
Renamed ALS_ENDPOINT_HOST to ALS_ENDPOINT

vijayg10:
Previously it was ALS_ENDPOINT. Somehow it is changed to ALS_ENDPOINT_HOST which is not referenced in any file other than config.js.
So reverting this name back to previous one.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -64,7 +64,7 @@ module.exports = {
         },
     },
     peerEndpoint: env.get('PEER_ENDPOINT').required().asString(),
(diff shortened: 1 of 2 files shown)


msk-:
In light of the following:
- the rename doesn't _appear_ to have been deliberate
- other similarly named variables don't have the `_HOST` suffix
- the chart uses the previous version (https://github.com/mojaloop/helm/blob/7a1708ecc6fcba013c7e3893c4490427613e357f/mojaloop-simulator/values.yaml#L291-L296)

I've approved.

PR #179, opened 2020-07-10
Changed async transistion lifecycle events to use Promsises.

kleyow:
Async life-cycle observer methods need to explicitly return a `Promise` in `javascript-state-machine` v3+. Without it transitions will stall the PSM.

This was very hard to diagnose :cry:

### DIFF
--- a/src/lib/model/common/PersistentStateMachine.js
+++ b/src/lib/model/common/PersistentStateMachine.js
@@ -11,23 +11,35 @@
 'use strict';
 const StateMachine = require('javascript-state-machine');
 
-async function saveToCache() {
+function saveToCache() {
     const { data, cache, key, logger } = this.context;
-    try {
-        const res = await cache.set(key, data);
-        logger.push({ res }).log(`Persisted model in cache: ${key}`);
-    }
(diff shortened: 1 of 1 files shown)


lewisdaly:
Hey @kleyow where are we at with this one? Did you and @eoln figure out what the problem was with the unit tests?

kleyow:
@lewisdaly @eoln Last update was that something with `saveToCache` was successfully failing in a sense and was halting the state machine for an unknown reason. We either need to fix the cache some how or have it moved out of the statemachine lifecycle functions.

Waiting to hear back from Pawel if he discovered anything while making some changes he was working on.

I'll follow up with him after I PR some sdk-standard-components stuff.

eoln:
@kleyow I removed `saveToCache` from the transition lifecycle. Now it should be called explicitly in `run` implementation when it is really needed, not on every transition change.

kleyow:
Yes, I saw, but had forgotten! I've resumed working on https://github.com/mojaloop/mojaloop/issues/314  
Closing this ticket since it's no longer needed.

PR #181, opened 2020-07-10
Feature/#1462 request for notif

vgenev:


src/InboundServer/api.yaml
```diff
@@ -1,4427 +1,4533 @@
-openapi: 3.0.0
+swagger: "2.0"
```
elnyry-sam-k, line 1: @vgenev - thanks for the PR; I think to help the reviews, we need to keep this to openapi 3 and just do the changes we need for v1.1 so that the changes are manageable

here's the openapi3 version of v1.1 FSPIOP API

src/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "10.6.1",
+  "version": "11.1.0",
```
elnyry-sam-k, line 3:
```suggestion
  "version": "11.0.0",
```

src/InboundServer/api.yaml
vijayg10: +1

bushjames:
@rmothilal @vgenev @elnyry-sam-k . Are there further PRs on this feature to come? I ask because it seems like there are outstanding things to deal with. Particularly, the inbound model does not appear to deal with an incoming PATCH notification yet?

vgenev:
@bushjames nope. no further PRs. The PATCH is handled just like a PUT in that case, so its the same handler atm: https://github.com/vgenev/sdk-scheme-adapter/blob/d9c0860c898f48afb23c12365aeeafdca228d01d/src/InboundServer/handlers.js#L975
it that's not the desired behaviour, please advice on the way forward.

bushjames:
thanks for the info @vgenev . I think it is wrong for the PATCH response to be handled by the putTransfersById handler. This handler is intended to only handle incoming fulfilments for outgoing transfers (when the scheme adapter is acting as payer). I think this actually needs a new handler which calls a new method on the inbound model to make a PUT /transfers/{id} call to the DFSP backend. This enables the DFSP backend to use the patch for the reason it is intended, i.e. to allow cashout in safety, knowing the switch has committed to give it the funds.

PR #185, opened 2020-07-14
Websocket API for test server

msk-:
### Motivation

Currently a vast range of tests that use the test server do something like this:
1. initiate some action in the switch
2. wait some predetermined time, `w`
3. using `/callbacks` or `/requests`, verify whether the action was completed as expected

The problem arises at (2). If `w` is too short, the test will likely fail. If `w` is too long, the test will take longer than necessary. This is exacerbated by environmental differences. To take two extremes: running on a 1000 processor, 1000TB cluster `w` will be very different to a cluster comprised of a handful of Raspberry Pis. The large cluster will wait _far longer_ than it needs to. The small cluster might attempt step (3) before the action has completed, causing a false negative test result.

Moreover, in order to define `w` for a given test `T`, one might run `T` multiple times to determine a good value for `u`. This leads to:
1. A _lot_ of effort to determine a good value for `u` for _every_ test, for _every_ cluster spec, multiplied over very many possible configurations.
2. An _expected failure rate_ corresponding to the chosen value of `u` and the system specification.

In other words, flaky tests.

This PR means that a test can set a high global test timeout config, say 60 seconds, and then wait for a notification that its event has occurred. This completely eliminates the problem highlighted above.

### Some potential points of implementation contention:
* Constants on Cache might be considered to be more of an application concern and less of a cache concern. Counterpoint: Cache is a Redis abstraction specifically for this application.
* Could explicitly pub/sub instead of using keyevent notifications. The change in this PR seemed like the smaller change, but is less philosophically pure.
* Configuration of redis keyevent notifications by the test server (when `ENABLE_TEST_FEATURES` is `true`).

### DIFF
--- a/src/TestServer/index.js
+++ b/src/TestServer/index.js
@@ -40,9 +40,26 @@ class TestServer {
     //    config set "notify-keyspace-events" "Eg$"
     //    See here: https://redis.io/topics/notifications#configuration
     //    Preferably only when test features are enabled
+    // 2. this._wsapi.on('error') below
 
     async setupApi() {
         this._api = new Koa();
+        this._wsapi = new ws.Server({ noServer: true });
+        this._wsClients = new Map();
+
+        this._wsapi.on('error', err => {
+            // TODO
+            console.log('ERROR');
+            throw err;
+        });
+
+        this._wsapi.on('connection', (socket, req) => {
+            this._wsClients.set(socket, req.url);
+            socket.on('close', () => {
+                this._wsClients.delete(socket)
+            });
+        });
+
         this._logger = await this._createLogger();
 
         this._cache = await this._createCache();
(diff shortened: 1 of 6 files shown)


sri-miriyala:
So do we need to update the postman scripts to subscribe for these responses from websocket instead of calling /sdktest/request and /sdktest/callback endpoints?

src/TestServer/index.js
```diff
@@ -127,8 +146,101 @@ class TestServer {
...
             server = http.createServer(this._api.callback());
         }
+
         return server;
     }
+
+    _createWsServer() {
+        const wss = new ws.Server({ noServer: true });
+
+        wss.on('error', err => {
```
bushjames, line 156: will this kill the entire server on any ws error? Why not "keep calm and carry on"?

src/TestServer/index.js
```diff
@@ -25,6 +26,10 @@ const router = require('@internal/router');
 const handlers = require('./handlers');
 const middlewares = require('../InboundServer/middlewares');
 
+const getWsIp = (req) => req.headers['x-forwarded-for']
+    ? req.headers['x-forwarded-for'].split(/\s*,\s*/)[0]
```
bushjames, line 30: Not sure what the purpose of this is? You appear to use it only in log entries...given that the x-forwarded-for list might not be in the correct order, always assuming the first entry seems a bit much of an assumption to make. I would have thought that the only important piece of information to log would be the IP of the ws peer itself.

src/TestServer/index.js
msk-: It will. I don't actually know what the error modes are. So I don't know if the server is likely to be operable after an error. I figured that this is likely to be run in a container in a container orchestration system, and therefore that
1. it should fail loudly so that someone knows it's failing
2. the container orchestration system will handle the "keep calm and carry on" bit

msk-: And that it will receive a PR if it needs to handle something specific.

src/TestServer/index.js
msk-: Good, thanks. Updated.

src/TestServer/index.js
bushjames: If it is the intention to actually kill the process at this point, setting process.exitCode will not do that by itself according to node docs. You need to do something like process.exit(1).

src/TestServer/index.js
bushjames: Have you pushed a commit? I dont see any change.

PR #187, opened 2020-07-22
fix: use anyOf instead of oneOf for PUT /authorizations/{id} at Inbou…

eoln:
- use anyOf instead of oneOf for PUT /authorizations/{id}  at InboundServer


lewisdaly:
@eoln do we want to bump the version here so we can publish a new docker image? Or are you just doing that in tags?

eoln:
@lewisdaly via tags

PR #188, opened 2020-07-22
fix: the way how errors are handled at InboundServer

eoln:
fix problems how the error are handled by InboudServer state machines

### DIFF
--- a/src/lib/model/InboundThirdpartyTransactionModel.js
+++ b/src/lib/model/InboundThirdpartyTransactionModel.js
@@ -82,13 +82,18 @@ class InboundThirdpartyTransactionModel {
         let mojaloopErrorCode = Errors.MojaloopApiErrorCodes.INTERNAL_SERVER_ERROR;
         if (err instanceof HTTPResponseError) {
             const e = err.getData();
-            if (e.res && e.res.body) {
-                try {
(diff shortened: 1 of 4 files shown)


src/lib/model/InboundThirdpartyTransactionModel.js
```diff
@@ -82,13 +82,18 @@ class InboundThirdpartyTransactionModel {
...
         if (err instanceof HTTPResponseError) {
             const e = err.getData();
-            if (e.res && e.res.body) {
-                try {
-                    const bodyObj = JSON.parse(e.res.body);
-                    mojaloopErrorCode = Errors.MojaloopApiErrorCodeFromCode(`${bodyObj.statusCode}`);
-                }
-                catch (ex) {
-                    this._logger.push({ ex }).log('Error parsing error message body as JSON');
+            if(e.res && (e.res.body || e.res.data)) {
```
kleyow, line 85: Does this mean that there's some mismatch some where in the construction of responses? I would of thought that there's some standard. 

Anyway I won't dig too deep into that. I'm fine with this.

eoln: @kleyow thanx!

PR #190, opened 2020-07-29
Added outbound GET/POST thirdparty requests transaction call.

kleyow:


kleyow:
Trying to get this checked in as is for now since we're moving over calls to the thirdparty-scheme-adapter.
Some talks with Lewis after stand-up.

- How will we be dealing with the interface with thirdparty scheme adapter. Since sdk-scheme-adapter has its own schemas for talking with DFSP's like how transferParty and Mojaloop `Party` are different. Summary was we will deal with it later/stick closer to Mojaloop's schema. That might change.

- Idea of multiple small state machines forming one larger one (my mind wandering idea). Decided against that. Stick to one scenario to one state machine.

- This is just a part of the outbound thirdparty transaction scenario that we will have to piece together later. https://github.com/mojaloop/pisp/blob/master/docs/out/state_diagrams/OutboundThirdpartyTransactionModel/OutboundThirdpartyTransactionModel.png but this is a good cutoff point since we're migrating to thirdparty-scheme-adapter.

PR #191, opened 2020-08-03
#1484: Update FSPIOP API version

oderayi:


src/InboundServer/api.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.0
 info:
-  version: '1.0'
+  version: '1.1'
   title: Open API for FSP Interoperability (FSPIOP)
   description: >-
     Based on API Definition.docx updated on 2018-03-13 Version 1.0. Note - The
```
elnyry-sam-k, line 6:
```suggestion
    Based on API Definition.docx updated on 2020-05-19 Version 1.1. Note - The API supports a maximum size of 65536 bytes (64 Kilobytes) in the HTTP header.
```

elnyry-sam-k:
Thanks @oderayi .. @bushjames , @msk- - any issues with upgrading to v1.1?

bushjames:
@elnyry-sam-k @oderayi From what I see, this PR just changes the inbound swagger version to 1.1? I do not see the point as this will have no effect on the functions of the adapter at all? Incoming requests will still be validated against resource 1.0 versions. No v.1.1 functionality is supported? There are no functional changes anywhere else?

oderayi:
Yes @bushjames. Some changes in line with the version change are coming in in a separate PR (https://github.com/mojaloop/sdk-scheme-adapter/pull/181)

elnyry-sam-k:
Thanks @oderayi , @bushjames - this is now included in #181 that steven referenced above, maybe that makes more sense. Thanks @vgenev

PR #192, opened 2020-08-07
fix(thirdparty): rename thirdParty to thirdparty

kleyow:


src/InboundServer/api.yaml
```diff
@@ -2604,15 +2604,15 @@ paths:
           $ref: '#/components/responses/ErrorResponse501'
         503:
           $ref: '#/components/responses/ErrorResponse503'
-  /thirdPartyRequest/transactions/{ID}/error:
+  /thirdpartyRequest/transactions/{ID}/error:
```
sridharvoruganti, line 2607: this should be /thirdpartyRequests/transactions/{ID}/error:

kleyow: done. thank you good spot.

PR #197, opened 2020-08-20
Enable test-integration job in pipeline

KamuelaFranco:


.circleci/config.yml
```diff
@@ -102,30 +102,30 @@ jobs:
...
-#          path: /home/circleci/project/junit.xml
+
+  test-integration:
+    executor: default-machine
+    steps:
+      - checkout
+      - run:
+          name: Install Docker Compose
+          command: |
+            curl -L https://github.com/docker/compose/releases/download/1.25.0/docker-compose-`uname -s`-`uname -m` > ~/docker-compose
```
bushjames, line 113: yuck. is there no "safe" way of installing docker-compose rather than curling a remote file onto the runner?

bushjames (changes requested):

.circleci/config.yml
KamuelaFranco: This was how it was working before. Absolve me of this sin. 😭

bushjames (approved):

PR #198, opened 2020-08-27
Add ISO-4217 test currencies to enable production testing scenarios.

bushjames:
Adding XTS and XXX test currencies to enable testing on production environments without using live currencies.


ksatya77 (dismissed):
LGTM

kirgene (approved):

PR #199, opened 2020-09-02
Feature/#1615 content headers

vgenev:
added config for resource API version


lewisdaly (changes requested):
Thanks @vgenev - just a couple minor points, and it would be nice to see some unit tests.

src/config.js
```diff
@@ -131,5 +131,7 @@ module.exports = {
     outboundErrorStatusCodeExtensionKey: env.get('OUTBOUND_ERROR_STATUSCODE_EXTENSION_KEY').asString(),
 
     proxyConfig: env.get('PROXY_CONFIG_PATH').asYamlConfig(),
-    reserveNotification: env.get('RESERVE_NOTIFICATION').default('false').asBool()
+    reserveNotification: env.get('RESERVE_NOTIFICATION').default('false').asBool(),
+    // resourcesVersion config should be string in format: "resouceOneName=1.0,resourceTwoName=1.1"
+    resourcesVersion: env.get('RESOURCES_VERSIONS').default('').asString()
```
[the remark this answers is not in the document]
vgenev: sure!

src/config.js
```diff
@@ -14,17 +14,46 @@ require('dotenv').config();
...
+            .reduce((p, c) => {
+                resourceVersionMap[p] = {
+                    contentVersion: c,
+                    acceptVersion: c.split('.')[0],
+                };
+            }));
+    return resourceVersionMap;
+}
+    
+function validateResourceVersions (resourceString) {
```
lewisdaly, line 42: I would rename this to `parseResourceVersions`, since it's both validating and parsing.

src/config.js
```diff
@@ -14,17 +14,46 @@ require('dotenv').config();
...
+    if (!resourceString) return '';
```
lewisdaly, line 43: Is this what we want to do? Won't this cause other downstream breaks in the `sdk-standard-components`? Perhaps we should return an empty map instead.

package-lock.json
```diff
@@ -0,0 +1,4178 @@
+{
```
lewisdaly, line 1: This file shouldn't be here.

lewisdaly (changes requested):
Just a few minor notes and thoughts.

src/config.js
vgenev: technically it won't break but makes much more sense to return empty map, yeah

package-lock.json
vgenev: that ba*d.. :>

src/test/unit/config.test.js
```diff
@@ -93,4 +93,23 @@ describe('config', () => {
...
+                acceptVersion: '1',
+                contentVersion: '1.1',
+            },
+
+        };
+        process.env.RESOURCE_VERSIONS = 'resourceOneName=1.0,resourceTwoName=1.1';
+        const config = require('../../config');
+        expect(config.resourceVersions).toEqual(resourceVersions);
+    });
+
```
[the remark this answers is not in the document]
vgenev: I was just following the pattern.. but yeah..

lewisdaly (approved):
Looks good!

PR #203, opened 2020-09-14
/#1653: Implement accented characters support

oderayi:


src/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "11.2.4",
+  "version": "11.2.5",
```
vgenev, line 3: I think in the new versioning model we need to change the version to 11.3.0, @elnyry-sam-k , @rmothilal can you confirm?

vgenev (changes requested):

src/package.json
elnyry-sam-k: yes that is correct. this is new non-breaking functionality, so 11.3.0

rmothilal (approved):
HAppy

bushjames (commented):
Nice work @oderayi !

PR #205, opened 2020-09-16
Bugfix/add transaction request service endpoint config

vijayg10:
**Bug:** When we try to send POST /transactionRequests to the sdk-scheme-adapter, its not able to send the callback.

**Fix:** Added `transactionRequestsEndpoint` and `TRANSACTION_REQUESTS_ENDPOINT` which are missing in the sdk-scheme-adapter. Where as sdk-standard-components library is expecting the config param 'transactionRequestsEndpoint' from sdk-scheme-adapter.

### DIFF
--- a/src/.env.example
+++ b/src/.env.example
@@ -56,14 +56,15 @@ CACHE_HOST=172.17.0.2
 CACHE_PORT=6379
 
 # SWITCH ENDPOINT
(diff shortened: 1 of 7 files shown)


lewisdaly (approved):
Looks good to me

PR #206, opened 2020-09-22
Feature/#1485 bulk quotes

vgenev:


src/lib/model/OutboundBulkTransfersModel.js
```diff
@@ -231,7 +231,7 @@ class OutboundBulkTransfersModel {
             bulkTransferId: this.data.bulkTransferId,
             bulkQuoteId: this.data.bulkQuoteId,
             payerFsp: this._dfspId,
-            payeeFsp: this.data.payeeFsp.fspId,
+            payeeFsp: this.data.individualTransfers[0].to.fspId,
```
ksatya77, line 234: @vgenev ... Do we have only one payeefsp across all the individual transfers?

vgenev: @ksatya77 in the switch the payee is only one of a bulk request. so the dfsp backend has to do the job to issue separate bulk requests to the switch for each different payee. it might well be in the sdk-scheme-adapter but its not in the current scope..

ksatya77 (approved):
Looks good

rmothilal (approved):
Happy

PR #224, opened 2020-11-04
Incorporated new sdk-standard-components functionality into SDK

msk-:
Two changes:

1. When a request fails with a 401, the new environment variable `WSO2_AUTH_FAILURE_REQUEST_RETRIES` will now allow the SDK to attempt to retrieve a new token, then retry the request, if WSO2 auth is enabled.
2. When WSO2 token retrieval fails, instead of failing silently, which has happened since the change of base request library from `request-promise` to `@mojaloop/sdk-standard-components/src/lib/request.js`, an error will now propagate to the top level (and crash the SDK)

### DIFF
--- a/src/InboundServer/handlers.js
+++ b/src/InboundServer/handlers.js
@@ -26,7 +26,7 @@ const getAuthorizationsById = async (ctx) => {
                 ...ctx.state.conf,
                 cache: ctx.state.cache,
                 logger: ctx.state.logger,
-                wso2Auth: ctx.state.wso2Auth,
+                wso2: ctx.state.wso2,
                 resourceVersions: ctx.resourceVersions,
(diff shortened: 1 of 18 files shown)


bushjames (changes requested):
I suggest this is a major version bump as it looks like default behaviour on auth failure has changed?

bushjames (approved):
I am happy that this does not change default behaviour on auth failure.

msk-:
> I am happy that this does not change default behaviour on auth failure.

Regarding this, James clarified with me on Slack:
1. the new environment variable is not mandatory
2. the default value of the new environment variable, when omitted, results in the same behaviour as before this change
3. there are no backward-incompatible API changes

PR #228, opened 2020-11-26
Added GET /quotes/{ID} to the inbound API

shashi165:
### Summary
Added the endpoint `GET /quotes/{ID}` to the inbound API. This will allow the payer to get the response  to a previous `POST /quotes` request. 
The `quoteResponse`  will be stored in the cache when `POST /quotes` is handled, and it will be retrieved on any subsequent  `GET /quotes/{ID}` and will be sent to the payer as a body of `PUT /quotes/{ID}` through switch. 
If no `quoteResponse`  is found in the cache then the `GET /quotes{ID}` will be ignored.
The payee is not supposed to request `GET /quotes{ID}`, if payee requests it, it will be ignored


src/lib/model/InboundTransfersModel.js
```diff
@@ -193,6 +193,9 @@ class InboundTransfersModel {
                 fulfilment: fulfilment
             });
 
+            // now store the quoteRespnse data against the quoteId in our cache to be sent as a response to GET /quotes/{ID}
```
mdebarros, line 196: Typo `quoteRespnse` -> `quoteResponse`

mdebarros (changes requested):
Minor typo. 

I think your approach needs to be changed.

I think it's ok to check the cache for a response, but I would rather recommend that you request the quote from the backend instead (i.e. the mojaloop-simulator). Or at least fall-back to the backend if the cache has expired.

I.e. the backend is really the true source here.

Thus I would recommend that a `this._backendRequests.getQuoteRequests(internalForm)` is implemented on the Mojaloop-Simulator, and that should be called instead.

src/lib/model/InboundTransfersModel.js
bushjames: I see both sides here; the originally sent quote response data has come from the backend so it should also respond to the GET request...but on the other hand, the scheme-adapter has it in its cache, so may as well use it and save making the call to the backend.

Scheme-adapter cache expiry seems to be the only reason to require a further call to the backend (the quote response itself CANNOT be different to that originally sent, can it?); but it could reasonably be assumed that the cache expiry period should be configured to hold the quote response until such time as it would otherwise have expired.

A good compromise would seem to be that the scheme-adapter return the cached value if it is in the cache; otherwise it should make a GET request to the backend. This requires adding a GET /quoterequests/{id} call to the DFSP backend API definition and, as Miguel mentions, implementing it in the simulator....I would strongly suggest/ask that we make this change backward compatible by not assuming the backend will implement this method. If an error is received from the backend when attempting this call some appropriate mojaloop error callback should be made to the caller.

bushjames (changes requested):
Looking awesome but please see my comment on Miguels comment.

src/lib/model/InboundTransfersModel.js
mdebarros: > A good compromise would seem to be that the scheme-adapter return the cached value if it is in the cache; otherwise it should make a GET request to the backend. This requires adding a GET /quoterequests/{id} call to the DFSP backend API definition and, as Miguel mentions, implementing it in the simulator....I would strongly suggest/ask that we make this change backward compatible by not assuming the backend will implement this method. If an error is received from the backend when attempting this call some appropriate mojaloop error callback should be made to the caller.

I am happy with this approach.

shashi165: @bushjames , I am trying to implement `GET /quoterequests/{quoteId}` in `mojaloop-simulator`. The response is stored in the `quote` table but ILP and condition is calculated in `sdk-scheme-adapter` while handling the `POST /quotes`, the backend has no idea about the these values, also the expiration time is added in `sdk-scheme-adapter`, and these values are sent in `PUT /quotes/{quoteId}` from `sdk-scheme-adapter`.
e.g.
```
{ 
   method: 'PUT', 
   uri: 'http://dev1-quoting-service.mojaloop.live/quotes/9c92c98b-12cc-4c07-a6e9-2362cfea5a66', 
   headers: { 
     'content-type': 'application/vnd.interoperability.quotes+json;version=1.0', 
     date: 'Tue, 01 Dec 2020 04:23:59 GMT', 
     'fspiop-source': 'payeefsp', 
     'fspiop-destination': 'payerfsp', 
     'fspiop-http-method': 'PUT', 
     'fspiop-uri': '/quotes/9c92c98b-12cc-4c07-a6e9-2362cfea5a66', 
     'fspiop-signature': '{"signature":"WMrwdUEU9w59hN0ODYW5_Bsl_yYRcRS9GUpzW31izJHYFUfXlZI6DTgKzLBqezAUUi29GIkY_Z0h0su4v317UEJWqj5ReM3Ff6-s9RCIGSSoFj18y1zAcfIfeSkcPW_aA8bl5W_qDVTXt5dyu6yZPXdhj97us__dc7-6O7W3ClltDcvP2Av2p7nC_fvgQbB_G7uGP4NmMc9ByeHS1zQm0C8h2UkO6yHk8ABJKU8GmLnxqC3lRjdBssRe6ICzanhdPsBpBuhpeYq_E1AgskG-5PAKmWkZyCOeTUqqTinb5xsU4OCouXXP2lnRHf5LyIGFpvmUVtGOgXzzk6pysPbokA","protectedHeader":"eyJhbGciOiJSUzI1NiIsIkZTUElPUC1VUkkiOiIvcXVvdGVzLzljOTJjOThiLTEyY2MtNGMwNy1hNmU5LTIzNjJjZmVhNWE2NiIsIkZTUElPUC1IVFRQLU1ldGhvZCI6IlBVVCIsIkZTUElPUC1Tb3VyY2UiOiJwYXllZWZzcCIsIkZTUElPUC1EZXN0aW5hdGlvbiI6InBheWVyZnNwIiwiRGF0ZSI6IlR1ZSwgMDEgRGVjIDIwMjAgMDQ6MjM6NTkgR01UIn0"}' 
   }, 
   body: '{"transferAmount":{"amount":"10","currency":"USD"},"expiration":"2020-12-01T04:24:59.755Z","ilpPacket":"AYIC9AAAAAAAAAPoHWcucGF5ZWVmc3AubXNpc2RuLjE3MDM5ODExOTA3ggLKZXlKMGNtRnVjMkZqZEdsdmJrbGtJam9pTmpJM01EbGtZVEF0TnpOa1l5MDBOV1k1TFRrM09HTXRNelExTldWaU9HWmhPR1V4SWl3aWNYVnZkR1ZKWkNJNklqbGpPVEpqT1RoaUxURXlZMk10TkdNd055MWhObVU1TFRJek5qSmpabVZoTldFMk5pSXNJbkJoZVdWbElqcDdJbkJoY25SNVNXUkpibVp2SWpwN0luQmhjblI1U1dSVWVYQmxJam9pVFZOSlUwUk9JaXdpY0dGeWRIbEpaR1Z1ZEdsbWFXVnlJam9pTVRjd016azRNVEU1TURjaUxDSm1jM0JKWkNJNkluQmhlV1ZsWm5Od0luMTlMQ0p3WVhsbGNpSTZleUp3WVhKMGVVbGtTVzVtYnlJNmV5SndZWEowZVVsa1ZIbHdaU0k2SWsxVFNWTkVUaUlzSW5CaGNuUjVTV1JsYm5ScFptbGxjaUk2SWpFM01ETTVPREV4T1RBM0lpd2labk53U1dRaU9pSndZWGxsY21aemNDSjlMQ0p3WlhKemIyNWhiRWx1Wm04aU9uc2lZMjl0Y0d4bGVFNWhiV1VpT25zaVptbHljM1JPWVcxbElqb2lUV0YwY3lJc0lteGhjM1JPWVcxbElqb2lTR0ZuYldGdUluMHNJbVJoZEdWUFprSnBjblJvSWpvaU1UazRNeTB4TUMweU5TSjlmU3dpWVcxdmRXNTBJanA3SW1GdGIzVnVkQ0k2SWpFd0lpd2lZM1Z5Y21WdVkza2lPaUpWVTBRaWZTd2lkSEpoYm5OaFkzUnBiMjVVZVhCbElqcDdJbk5qWlc1aGNtbHZJam9pVkZKQlRsTkdSVklpTENKcGJtbDBhV0YwYjNJaU9pSlFRVmxGVWlJc0ltbHVhWFJwWVhSdmNsUjVjR1VpT2lKRFQwNVRWVTFGVWlKOWZRAA","condition":"fO_4g_pLRbST9ax6raKWY2OSapcFXrBkl7-dI1PCUME","payeeFspFee":{"amount":"0","currency":"USD"},"payeeFspCommission":{"amount":"0","currency":"USD"}}'
}```

How can I address the problem of ILP, condition and expiration not being available in the backend for generating the response to `GET /quotesrequest/{id}`. Any ideas?

bushjames: Good question @shashi165 . It is true that some of the quote response values are calculated in the scheme adapter. This would seem to force us down the route of returning a 404 if the cache has expired. Although not quite as elegant, I think this is acceptable given that the quote can realistically be assumed to have a lifetime similar to that of the cache expiry, at least in order of magnitide. What do you think @mdebarros ?

mdebarros: @bushjames I believe we should validate/invalidate the following assumption before we can assume the cache approach is acceptable:
-  GET /Quote request should always return a quote that was processed by the system

However, we do have two alternatives here:
1. we could also make the cache a persistent store, thereby we do not need to worry about the data expiring from the cache. <---This would actually allow the FSP to decide the approach in handling the GET /quotes since they can choose to deploy the Redis store as a cache or a persistent store.
2. we could calculate the transfer time before the call to the backend, and include that as part of the call to the backend. The caveat here is that the ENV var that is used to calculate the timeout must be modified to compensate for the latency of the backend call.

Let me know your thoughts.

bushjames: Indeed redis can be used as a persistent store. The defaults used by the scheme adapter set no expiry at all, leaving the user to configure store wide expiry should they so wish. I do think though, that it is OK to just return a 404 if the cache has expired. I cannot imagine a user setting the cache expiry deliberately shorter than their quote and/or transfer expiry. Although it would be possible to do so, it would not be sensible IMO. Perhaps a few words to that effect in the README is sufficient, and some comments in the code where we would send a 404?

mdebarros: > Indeed redis can be used as a persistent store. The defaults used by the scheme adapter set no expiry at all, leaving the user to configure store wide expiry should they so wish. I do think though, that it is OK to just return a 404 if the cache has expired. I cannot imagine a user setting the cache expiry deliberately shorter than their quote and/or transfer expiry. Although it would be possible to do so, it would not be sensible IMO. Perhaps a few words to that effect in the README is sufficient, and some comments in the code where we would send a 404?

What are your thoughts on the 2nd alternative option?

2. we could calculate the transfer time before the call to the backend, and include that as part of the call to the backend. The caveat here is that the ENV var that is used to calculate the timeout must be modified to compensate for the latency of the backend call.

bushjames: @mdebarros I think the latency of the backend would be too unpredictable for us to make a sensible judgement when calculating the expiry.

bushjames (approved):
Looking good Shashi.

mdebarros (commented):
+1

mdebarros (approved):

PR #237, opened 2020-12-31
chore: sync pisp/master to v11.10.1

kleyow:


src/lib/model/OutboundAuthorizationsModel.js
```diff
@@ -132,31 +132,31 @@ async function onRequestAuthorization() {
...
         try {
-            // in InboundServer/handlers is implemented putAuthorizationsById handler 
+            // in InboundServer/handlers is implemented putAuthorizationsById handler
             // where this event is fired but only if env ENABLE_PISP_MODE=true
             subId = await cache.subscribe(channel, async (channel, message, sid) => {
-                try { 
+                try {
                     const parsed = JSON.parse(message);
                     this.context.data = {
                         ...parsed,
```
kleyow, line 141: @eoln is this supposed to be `...parsed` or `...parsed.data`?
It doesn't line up with the test found at 
https://github.com/mojaloop/sdk-scheme-adapter/blob/734e0fae8e17fa0c24b2fd1c79d2880915c4c247/src/test/unit/lib/model/OutboundAuthorizationsModel.test.js#L146
https://github.com/mojaloop/sdk-scheme-adapter/blob/734e0fae8e17fa0c24b2fd1c79d2880915c4c247/src/test/unit/lib/model/OutboundAuthorizationsModel.test.js#L167 

I'm not sure how we want to store the message or if it's the test that's faulty. It's creating a false positive since the `expect(...)` is run after an asyncronyous function. (Which is an issue of itself, feel like that was discussed at one point)

I vote we follow up on this maybe in a different PR since it's a little out of context for this merge ticket.

eoln (approved):
@kleyow I suggest fixing this false positive test in separate PR, at the end of merging with the latest `master` branch. If there will be no time, or there will be unexpected complications,  please create a separate story ticket to expose this hidden work.

kleyow:
Will do.

PR #238, opened 2021-01-04
chore: update config in models

kleyow:


lewisdaly (approved):
It surprises me that this doesn't break any tests...

kleyow:
Yes because the tests actually point to the right config haha, so the errors won't show until this is deployed.

PR #239, opened 2021-01-04
fix: edit some faulty tests

kleyow:
Yes you were correct Pawel. `then` runs after the test is finished.
- Changed to `parsed.data`
- Moved code out of `then` block

### DIFF
--- a/src/test/unit/lib/model/common/PersistentStateMachine.test.js
+++ b/src/test/unit/lib/model/common/PersistentStateMachine.test.js
@@ -169,8 +169,7 @@ describe('PersistentStateMachine', () => {
             const psm = await PSM.create(data, cache, key, logger, smSpec);
             checkPSMLayout(psm);
 
-            // transition `init` should encounter exception when saving `context.data`
-            expect(() => psm.init()).rejects.toEqual(new Error('error from cache.set'));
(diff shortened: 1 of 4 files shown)


src/test/unit/lib/model/common/PersistentStateMachine.test.js
```diff
@@ -169,8 +169,7 @@ describe('PersistentStateMachine', () => {
             const psm = await PSM.create(data, cache, key, logger, smSpec);
             checkPSMLayout(psm);
 
-            // transition `init` should encounter exception when saving `context.data`
-            expect(() => psm.init()).rejects.toEqual(new Error('error from cache.set'));
+            expect(() => psm.saveToCache()).rejects.toEqual(new Error('error from cache.set'));
```
kleyow, line 172: Don't believe `init` calls `cache.set` at all so it never throws the error. Maybe `saveToCache` was called in a previous lifecycle method, not sure...Changed it to just call the method directly.

eoln: we dropped auotmatic `cache.set` from `onAfterTransition` https://github.com/mojaloop/sdk-scheme-adapter/blob/pisp/master/src/lib/model/common/PersistentStateMachine.js#L26-L30

eoln (approved):

PR #243, opened 2021-01-26
feat: 1991 get parties

eoln:
Implementing get /parties @ Outbound interface

### DIFF
--- a/src/InboundServer/api.yaml
+++ b/src/InboundServer/api.yaml
@@ -686,6 +686,13 @@ paths:
                 middleName: Pierre
                 lastName: Trudeau
               dateOfBirth: '1971-12-25'
+            accounts:
+            - currency: USD
+              description: savings
+              address: moja.red.8f027046-b82a-4fa9-838b-70210fcf8136
+            - currency: USD
+              description: checkings
+              address: moja.red.8f027046-b82a-4fa9-838b-70210fcf8137
       responses:
         '200':
           $ref: '#/components/responses/Response200'
--- a/src/OutboundServer/api.yaml
+++ b/src/OutboundServer/api.yaml
@@ -356,6 +356,47 @@ paths:
         504:
           $ref: '#/components/responses/accountsCreationTimeout'
 
+  /parties/{Type}/{ID}:
+    parameters:
+      - $ref: '#/components/parameters/Type'
+      - $ref: '#/components/parameters/ID'
+    get:
+      description: >-
+        The HTTP request GET /parties// (or GET /parties///) is used to lookup
+        information regarding the requested Party, defined by ,  and optionally
+        (for example, GET /parties/MSISDN/123456789, or GET
+        /parties/BUSINESS/shoecompany/employee1).
+      summary: PartiesByTypeAndID
+      tags:
+        - parties
+      operationId: PartiesByTypeAndID
+      responses:
+        '200':
+          $ref: '#/components/responses/PartiesByIdResponse'
+        '404':
+          $ref: '#/components/responses/PartiesByIdError404'
+
+  /parties/{Type}/{ID}/{SubId}:
+    parameters:
+      - $ref: '#/components/parameters/Type'
+      - $ref: '#/components/parameters/ID'
+      - $ref: '#/components/parameters/SubId'
+    get:
+      description: >-
+        The HTTP request GET /parties// (or GET /parties///) is used to lookup
+        information regarding the requested Party, defined by ,  and optionally
+        (for example, GET /parties/MSISDN/123456789, or GET
+        /parties/BUSINESS/shoecompany/employee1).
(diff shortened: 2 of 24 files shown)


lewisdaly (dismissed):

src/lib/model/PartiesModel.js
```diff
@@ -0,0 +1,257 @@
...
+
+
+/**
+ * @name create
+ * @description creates a new instance of state machine specified in specStateMachine ^
+ * @param {Object} data     - payload data
+ * @param {String} key      - the cache key where state machine will store the payload data after each transition
+ * @param {Object} config   - the additional configuration for transition handlers
+ */
+async function create(data, key, config) {
```
bushjames, line 230: I see you have gone with a more "functional" paradigm for this model, which is not in itself a problem. I am a little concerned that this would introduce a different pattern from the rest of the codebase as the existing models use a "class" pattern. I think it would be best to either stick to the existing pattern in the other models or refactor those to follow this pattern, although I realise that would be significant volume of work. I think it is not super fantastic to have the models implemented in two different patterns.

bushjames (changes requested):

src/OutboundServer/api.yaml
```diff
@@ -1630,6 +1671,518 @@ components:
...
+          description: 'Display name of the Party, could be a real name or a nick name.'
+        personalInfo:
+          $ref: '#/components/schemas/PartyPersonalInfo'
+          description: >-
+            Personal information used to verify identity of Party such as first,
+            middle, last name and date of birth.
+      required:
+        - partyIdInfo
+
+    PartyName:
```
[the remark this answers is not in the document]
eoln: I was thinking, that may be using @mojaloop/api-snippets would be a better solution - like here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/interface/api-outbound-template.yaml#L473

src/lib/model/PartiesModel.js
eoln: I agree that having two patterns for models brings some inconsistency. I believe that this `functional` proposition could bring cleaner code. I didn't start refactoring all other models, because first I want to know your opinion about this approach. So what do you favor: `class'ish` or the `functional` approach to structure the module? (of course in javascript `class` is a `function` :-D

bushjames: @eoln I do not have a particularly strong opinion either way, but maybe because of my background with OO languages I personally slightly favour the "class" oriented approach. I am aware though that encapsulating state can have some consequences which complicate testing, like having to mock many dependencies and internal state initialisation etc... which brings its own issues...if functional style is well implemented it can eliminate some of that, but also if some basics are done well in OO approach those negatives can be overcome also. In the mojaloop ecosystem I think we have to both judge the opinion of "the community" but also offer some guidance as to what is sensible. Your code looks clean, elegant, understandable, extensible and maintainable and I would personally be happy if the other models followed this pattern also.

src/OutboundServer/api.yaml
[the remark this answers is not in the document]
bushjames: I like this approach @eoln . By all means go with snippets here also.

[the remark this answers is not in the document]
eoln: @bushjames I would like to go with snippets in separate PR, elsewhere it will be a really big PR

src/lib/model/PartiesModel.js
eoln: @bushjames we have a plan to create more sync api endpoints @ Outgoing interface (at least three more:  GET /quotes, POST /authorizations, POST /transfer) so they can be called by thirdparty-scheme-adapter (scheme adapters tandem).  So we have to make a decision @lewisdaly: should we think about refactoring other already existing models in sdk-scheme-adapter, or simply refactor only GET /parties to use the existing approach...

src/OutboundServer/api.yaml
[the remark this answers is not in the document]
eoln: we even have a story for this: https://app.zenhub.com/workspaces/mojaloop-project-59edee71d1407922110cf083/issues/mojaloop/project/1998

bushjames (approved):

src/InboundServer/api.yaml
```diff
@@ -686,6 +686,13 @@ paths:
                 middleName: Pierre
                 lastName: Trudeau
               dateOfBirth: '1971-12-25'
+            accounts:
```
kleyow, line 689: This is pisp specific. Don't think this should be going into  `master` as of yet

eoln: but this property is optional

kleyow: Sure that's fine. Don't let it hills back these changes. +1

PR #246, opened 2021-02-04
feat: 1991 get parties timeout

eoln:
introduce of `defferedJob` module to handle in a consistent way async to sync the call transformation


src/lib/model/PartiesModel.js
```diff
@@ -124,43 +125,23 @@ function getResponse() {
...
-                    };
-                    resolve();
-                } catch(err) {
-                    reject(err); 
-                }
-            });
-            
-            // GET /parties request to the switch
+    
+    return deferredJob(cache, channelName(type, id, subId))
```
kleyow, line 131: This reads a lot nicer.

kleyow (approved):

PR #257, opened 2021-02-23
chore: publish-npm

eoln:


sridharvoruganti (approved):
lgtm

partiallyordered:
Looks fine, but why does it need a version bump?

partiallyordered (approved):

eoln:
patch version should be bumped for maintenance (chore) changes

partiallyordered:
Ah, there's already an `11.4.0` release. This makes sense.
Are we bumping patch versions for maintenance changes as a precaution against unforeseen consequences? E.g. would we bump versions for a README change to guard against the possibility that the README is somehow incorporated into a produced artifact and we weren't aware of it?

eoln:
README is part of the documentation. In most cases, we want to have a doc in any deployment artifact. It is not breaking change, so the patch should be bumped. 

IMHO: `For any PR we should always bump version`

partiallyordered:
Okay what I was really asking was about this:

> For any PR we should always bump version

I was trying to ask: what's the justification for that?

PR #258, opened 2021-02-23
refactor: drop using internal modules

eoln:
Using `file:` when specifying the dependencies unable to use sdk-scheme-adapter as npm module -> the `npm install` fails with ENOLOCAL. 
According to npm documentation `file:` feature shouldn't be used when we want to make module public in npm repository: 
https://docs.npmjs.com/cli/v7/configuring-npm/package-json
```This feature is helpful for local offline development and creating tests that require npm installing where you don't want to hit an external server, but should not be used when publishing packages to the public registry.```


sridharvoruganti (dismissed):
lgtm

bushjames (dismissed):
I see no issues. Many thanks @eoln

lewisdaly:
__please don't merge!__ @kleyow  is fixing merge issues

kleyow (approved):

PR #278, opened 2021-06-17
feat(#2264): add more robust header validation for inbound server

kleyow:
The real solution (:thinking:) is to refactor the inbound server to Hapi and use `central-services-shared`'s header validator...

I'm assuming https://github.com/mojaloop/project/issues/2264 doesn't want that big of a code refactor. So I've ripped the header validator code and inserted it into the Inbound middleware.

lmk if this matches the acceptance criteria @mdebarros


mdebarros (approved):
LGTM

PR #279, opened 2021-07-27
fix(inbound-api): fixes for extension list and tls config

oderayi:


src/InboundServer/middlewares.js
```diff
@@ -331,11 +331,12 @@ const createJwsValidator = (logger, keys, exclusions) => {
  * @return {Function}
  */
 const applyState = (sharedState) => async (ctx, next) => {
+    const outboundRequests = (sharedState.conf && sharedState.conf.mutualTLS && sharedState.conf.mutualTLS.outboundRequests) || {};
     Object.assign(ctx.state, {
         ...sharedState,
         conf: {
             ...sharedState.conf,
-            tls: sharedState?.conf?.mutualTLS?.outboundRequests,
+            tls: outboundRequests,
```
bushjames, line 339: why this change?

bushjames (changes requested):

src/InboundServer/middlewares.js
oderayi: @bushjames The previous construct was throwing error as it appears not be valid JavaScript.

bushjames (approved):
I cannot spot any issues.

PR #280, opened 2021-08-04
feat: added ilp packet to transfer object

shashi165:


bushjames (approved):
I cannot spot any issues.

PR #287, opened 2021-10-20
fix: adding currencies xxx and xts which are part of iso 4217

sivachundur:


kleyow:
@lewisdaly tagging you, since this is cleaning up the `/authorization` endpoint as a side product that used to be in the thirdparty transfer flow.

lewisdaly:
> @lewisdaly tagging you, since this is cleaning up the `/authorization` endpoint as a side product that used to be in the thirdparty transfer flow.

Thanks @kleyow  - all good, we aren't using that endpoint anymore in the sdk-scheme-adapter.

PR #320, opened 2022-05-30
feat!: added outbound bulk api

shashi165:
BREAKING CHANGE:
The bulk transfers api has been re-defined, hence old bulkTransfers requests won't work anymore


PaulGregoryBaker (commented):
The PATCH need to change to a PUT.

Sam raised the point that only Mojaloop can raise a Patch notification, and Patch should be a notification and not change or affect anything at the switch.

src/OutboundServer/api.yaml
```diff
@@ -165,33 +165,37 @@ paths:
...
           $ref: '#/components/responses/bulkTransferBadRequest'
         '500':
-          $ref: '#/components/responses/bulkTransferServerError'
-        '504':
-          $ref: '#/components/responses/bulkTransferTimeout'
+          $ref: '#/components/schemas/errorResponse'
   '/bulkTransfers/{bulkTransferId}':
-    get:
-      summary: Retrieves information for a specific bulk transfer
+    patch:
```
PaulGregoryBaker, line 175: The PATCH needs to change to a PUT.

Sam raised the point that only Mojaloop should raise patch notification, and a Patch should only be a notification. I.e. it should not affect a change at the hub.

elnyry-sam-k (approved):

mdebarros (commented):
It looks good to me based on the latest [design](https://github.com/mojaloop/documentation/pull/371).

However, is this API definition not generated from API snippets?

src/OutboundServer/api.yaml
```diff
@@ -470,6 +474,37 @@ paths:
...
+      type: object
+      required:
+        - enabled
+      properties:
+        enabled:
+          type: boolean
+          enum:
+            - false
+            - true
+    autoAcceptQuoteFalse:
```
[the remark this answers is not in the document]
shashi165: when `autoAcceptQuote` is enabled then `perTransferFeeLimits` is mandatory 
@PaulGregoryBaker am I correct?

elnyry-sam-k:
@shashi165 - can you also please change the version to 2.0.0 - this being a major change because of new resource addition

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "14.0.0",
+  "version": "14.1.0",
```
mdebarros, line 3: This is not necessary, the auto-release process will bump this up.

Also, since we have a new major version for the API, do we not need to make this a Major release as well?

mdebarros (changes requested):

package.json
shashi165: done

src/OutboundServer/api.yaml
```diff
@@ -484,25 +484,17 @@ components:
...
-        enabled:
-          type: boolean
-          enum:
-            - false
-    autoAcceptQuoteTrue:
-      type: object
-      required:
-        - enabled
-        - perTransferFeeLimit
+        - perTransferFeeLimits
```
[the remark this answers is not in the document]
shashi165: yes

PR #324, opened 2022-06-29
mbp-633-638 Add interface, types, errors and aggregate for lookup phase

gibaros:


modules/command-handler/src/application/aggregate.ts
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	async destroy(): Promise<void> {
+		await this.repo.destroy();
+	}
+
+     // TODO: Add update
+ 
+     //BulkTransferEntity
+     async createBulkTransferEntity(BulkTransferEntity: BulkTransferEntity): Promise<string> { // TODO: BulkTransferEntity or IBulkTransferEntity?
```
mdebarros, line 73: These functions are probably going to match most of the Commands that appear on the Detail Design.

Nothing wrong with having these, but they may not end up being very useful since allot of the "crud" like operations will be handled by the Repo.

modules/command-handler/src/application/infrastructure-interfaces/irepo.ts
```diff
@@ -0,0 +1,66 @@
...
+ - Juan Correa <juancorrea@modusbox.com>
+
+ --------------
+ ******/
+
+"use strict";
+
+import {IBulkTransferEntity, BulkTransferRequest, IndividualTransfers, BulkBatch} from "@mojaloop/sdk-scheme-adapter-private-types"; // TODO: mbp-638
+
+export interface IBulkTransferEntityRepo {
```
mdebarros, line 35: Looks good.

@vijayg10 do you see anything missing here?

modules/command-handler/src/application/infrastructure-interfaces/irepo.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Modusbox
+ - Shashikant Hirugade <shashikant.hirugade@modusbox.com>
+ - Juan Correa <juancorrea@modusbox.com>
+
+ --------------
+ ******/
+
+"use strict";
+
+import {IBulkTransferEntity, BulkTransferRequest, IndividualTransfers, BulkBatch} from "@mojaloop/sdk-scheme-adapter-private-types"; // TODO: mbp-638
```
mdebarros, line 33: I would recommend adding "Entity" to all Entities.

```suggestion
import {IBulkTransferEntity, BulkTransferRequestEntity, IndividualTransfersEntity, BulkBatchEntity} from "@mojaloop/sdk-scheme-adapter-private-types"; // TODO: mbp-638
```

modules/command-handler/src/application/types.ts
```diff
@@ -0,0 +1,749 @@
...
+    InvalidBulkTransferEntityPartyLookupTotalCountTypeError,
+    InvalidBulkTransferEntityPartyLookupSuccessCountTypeError,
+    InvalidBulkTransferEntityPartyLookupFailedCountTypeError,
+    InvalidBulkTransferEntityBulkQuotesTotalCountTypeError,
+    InvalidBulkTransferEntityBulkQuotesFailCountTypeError,
+    InvalidBulkTransferEntityBulkTransfersTotalCountTypeError,
+    InvalidBulkTransferEntityBulkTransfersSuccessCountTypeError,
+    InvalidBulkTransferEntityBulkTransfersFailCountTypeError
+} from "./errors";
+
```
mdebarros, line 71: I am not sure these should be in a "types" file, as they are concrete class implementations of types/interfaces.

I feel like these files should instead be called `entity.ts` or something similar.

modules/command-handler/src/application/types.ts
```diff
@@ -0,0 +1,749 @@
...
+    partyLookupSuccessCount: number;
+    partyLookupFailedCount: number;
+    bulkQuotesTotalCount: number;
+    bulkQuotesSuccessCount: number;
+    bulkQuotesFailedCount: number;
+    bulkTransfersTotalCount: number;
+    bulkTransfersSuccessCount: number;
+    bulkTransfersFailedCount: number;
+
+    constructor(
```
mdebarros, line 89: You could also just pass the IBulkTransferEntity interface in the constructor.

But lets chat about that on a call.

modules/private-types/src/types.ts
```diff
@@ -0,0 +1,336 @@
...
+// TODO: check name and status enums.
+export enum IBulkTransferStatus {
+	ACTIVE = "RECEIVED",
+	DELETED = "DISCOVERY_PROCESSING",
+	AGREEMENT_PROCESSING = "AGREEMENT_PROCESSING",
+	TRANSFER_PROCESSING = "TRANSFER_PROCESSING"
+}
+
+export enum IIdType {
+	MSISDN = "MSISDN",
```
mdebarros, line 164: Why only these types? 

All types that are allowable should be adhering to the FSPIOP Specification.

Another option is to not make this specific.

mdebarros (commented):
Some comments for your attention.

@vijayg10 please also review.

modules/command-handler/src/application/infrastructure-interfaces/irepo.ts
vijayg10: I think we should create Repo for aggregates, not for entities. Isn't it?

vijayg10 (approved):

PR #329, opened 2022-07-04
fix: creating inbound server

kirgene:
Fix creating InboundServer object

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,7 +78,6 @@ class Server extends EventEmitter {
             this.conf,
             this.logger.push(LOG_ID.INBOUND),
             this.cache,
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -78,7 +78,6 @@ class Server extends EventEmitter {
             this.conf,
             this.logger.push(LOG_ID.INBOUND),
             this.cache,
-            this.metricsClient,
```
kleyow, line 81: Is it ok to remove this? What was it doing?

kirgene: that's a copy-paste type

mdebarros (approved):
+1

PR #344, opened 2022-08-12
fix(mojaloop/2886): fix bulk-quotes and bulk-transfers functionality

vijayg10:


docker/ml-testing-toolkit/spec_files/system_config.json
```diff
@@ -9,31 +11,52 @@
...
+    "ENABLED": false,
     "API_URL": "http://connection-manager-api:5061",
     "AUTH_ENABLED": false,
     "HUB_USERNAME": "hub",
     "HUB_PASSWORD": "hub"
   },
   "KEYCLOAK": {
     "ENABLED": false,
-    "API_URL": "http://172.17.0.1:9080",
+    "API_URL": "http://localhost:8080",
```
mdebarros, line 36: I dont think this will work on mac.

Can you not use the service name instead?

Or are you trying to hit a service locally?

vijayg10: HI Miguel, these are just dummy values.

src/InboundServer/handlers.js
```diff
@@ -735,7 +735,7 @@ const postBulkQuotes = async (ctx) => {
  */
 const putBulkQuotesById = async (ctx) => {
     // publish an event onto the cache for subscribers to action
-    await ctx.state.cache.publish(`bulkQuotes_${ctx.state.path.params.ID}`, {
+    await ctx.state.cache.publish(`bulkQuote_${ctx.state.path.params.ID}`, {
```
mdebarros, line 738: Why are we changing this?

And would it impact an existing deployment since we are changing the keys here?

*the Same comment applies elsewhere through the PR.

src/lib/model/OutboundBulkTransfersModel.js
```diff
@@ -15,7 +15,7 @@ const { uuid } = require('uuidv4');
 const StateMachine = require('javascript-state-machine');
 const { MojaloopRequests } = require('@mojaloop/sdk-standard-components');
 const { BackendError } = require('./common');
-
+const shared = require('./lib/shared');
```
mdebarros, line 18: We should really start using aliases throughout our code

Example package.json config: https://github.com/mojaloop/central-ledger/blob/master/package.json#L29
Example usage: https://github.com/mojaloop/central-ledger/blob/336a0a27e908eedeb0dcf8b171ad8c0edfb4c3d8/test/unit/handlers/bulk/shared/validator.test.js#L7

mdebarros (commented):
No major issues, just some comments for you to respond too.

src/InboundServer/handlers.js
vijayg10: This is a bug. The outbound handler subscribed to bulkQuote where as inbound handler is publishing to 'bulkQuotes' which is wrong channel.
As per this line https://github.com/mojaloop/sdk-scheme-adapter/blob/37b1b389a314e9cd53a4a8785a44d48547b6e920/src/InboundServer/handlers.js#L842
I changed that to right value and also it follows the same pattern as transfers.

It seems nobody is using bulk functionality of scheme adopter with all these broken things

src/lib/model/OutboundBulkTransfersModel.js
vijayg10: Sure, there are lot of places. May be we need a new story or we can do that as part of the new bulk implementation.

mdebarros (approved):
+1

PR #348, opened 2022-08-22
running only the integration tests that pass without bugs

sri-miriyala:


modules/outbound-command-event-handler/test/integration/application/outbound_command_event_handler.test.ts
```diff
@@ -827,7 +827,7 @@ describe("Tests for Outbound Command Event Handler", () => {
 
   // // TESTS FOR QUOTE PROCESSING
 
-  // test("When Inbound command event ProcessSDKOutboundBulkQuotesRequest is received\
```
mdebarros, line 830: The above changes look good, but why do we have all these tests commented out below?

sri-miriyala: For this sprint in the story I am working on, I am only adding the test cases not the test implementations. So I am just adding the description. In my latest commit, I uncommented and am using test.skip to skip the tests that should not be running yet. @mdebarros

mdebarros: Ok. Sounds fine, however if you are going to add something commented-out, then please include a comment above it with a short description or `TODO: <something>`.

mdebarros (approved):
+1

PR #354, opened 2022-08-25
refactor(mojaloop/#2883): move repo code out of command module and into shared lib

kleyow:
refactor(mojaloop/#2883): move repo code out of command module and into shared lib - [https://github.com/mojaloop/project/issues/2883](https://github.com/mojaloop/project/issues/2883)

- Move repo code out of command module and into shared lib
- Add redis config for domain module


modules/private-shared-lib/src/domain/bulk_transaction_agg/index.ts
```diff
@@ -167,23 +165,4 @@ export class BulkTransactionAgg extends BaseAggregate<BulkTransactionEntity, Bul
             // this._partyLookupFailedCount = undefined;
         }
     }
-
-
-    static async ProcessCommandEvent(
```
mdebarros, line 172: Why was this removed?

mdebarros (changes requested):
One issue for you to look at, and also I would suggest taking advantage of the Alias's defined n the tsconfig for imports: Ref: https://github.com/kleyow/sdk-scheme-adapter/blob/feat/PartyInfoCallbackProcessed/modules/outbound-command-event-handler/tsconfig.json#L7

I don't think we have defined similar alias's in other modules, but we should.

modules/private-shared-lib/src/domain/bulk_transaction_agg/index.ts
kleyow: It has been moved here https://github.com/mojaloop/sdk-scheme-adapter/pull/354/files#diff-7d1faa4862c22f5fcf174cadd8101d3cf5a0ede1e347c600ec78bd2758340bb5R92 I saw no reason it was a static method on this class so I moved it to the handler.

mdebarros: `bulk_transaction_agg` <-- SHOULD only exist in the Command handler.

I believe you were ONLY suppose to move the REPO, NOT Aggregates!

kleyow:
closed infavor of org branch https://github.com/mojaloop/sdk-scheme-adapter/pull/358

PR #361, opened 2022-09-01
fix: redis timeout while running integration tests

kirgene:
fix redis timeout while running integration tests


modules/outbound-command-event-handler/test/integration/application/outbound_command_event_handler.test.ts
```diff
@@ -73,16 +73,19 @@ const bulkTransactionEntityRepo = new RedisBulkTransactionStateRepo(bulkTransact
...
 
 describe("Tests for Outbound Command Event Handler", () => {
-
   beforeEach(async () => {
+    domainEvents = [];
+  });
+
+  beforeAll(async () => {
     await producer.init();
     domainEvents = [];
```
mdebarros, line 82:
```suggestion
```

Already handled by the `beforeEach`.

mdebarros (changes requested):
+1

.circleci/config.yml
```diff
@@ -205,7 +205,17 @@ jobs:
       - run:
           name: Execute Integration Tests
           command: |
-            yarn run test:integration
+            run_int_tests() {
```
mdebarros, line 208: Why are we not using docker-compose to start up the handlers similar to how we do so with the api-svc?

I believe everything should work as expected if we used the existing mechanism.

mdebarros (changes requested):

.circleci/config.yml
kirgene: running all services altogether breaks integration tests.
services start to communicate with each other, and test consumers get only a portion of all messages

mdebarros (approved):
+1

PR #363, opened 2022-09-06
docs: openapi bulk doc update

kirgene:
* add PlantUML API sequence diagram (from mojaloop/documentation)
* update DFSP inbound OpenAPI docs


mdebarros (approved):
+1

Although going to rely on @PaulGregoryBaker to double confirm the SD.

PR #374, opened 2022-09-14
fix(mojaloop/2809): individual transfer results is empty

vijayg10:
fix([mojaloop/2809](https://github.com/mojaloop/project/issues/2809)): individual transfer results is empty

### DIFF
--- a/modules/api-svc/src/FSPIOPEventHandler/handlers/party-info-requested.js
+++ b/modules/api-svc/src/FSPIOPEventHandler/handlers/party-info-requested.js
@@ -56,7 +56,11 @@ module.exports.handlePartyInfoRequestedDmEvt = async (
             bulkId: event.getKey(),
             content: {
                 transferId: event.getTransferId(),
-                partyResult: response,
+                partyResult: {
+                    party: response.party?.body,
+                    currentState: response.currentState,
+                    errorInformation: response.errorInformation
+                },
             },
             timestamp: Date.now(),
             headers: [],
--- a/modules/outbound-command-event-handler/package.json
+++ b/modules/outbound-command-event-handler/package.json
@@ -53,9 +53,9 @@
   },
   "devDependencies": {
     "@types/convict": "^6.1.1",
-    "@types/express": "^4.17.13",
+    "@types/express": "^4.17.14",
     "@types/jest": "^29.0.2",
-    "@types/node": "^18.7.17",
+    "@types/node": "^18.7.18",
     "@types/node-cache": "^4.2.5",
     "@types/supertest": "^2.0.12",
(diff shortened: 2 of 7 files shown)


kleyow (approved):

modules/api-svc/src/FSPIOPEventHandler/handlers/party-info-requested.js
```diff
@@ -56,7 +56,11 @@ module.exports.handlePartyInfoRequestedDmEvt = async (
             bulkId: event.getKey(),
             content: {
                 transferId: event.getTransferId(),
-                partyResult: response,
```
kirgene, line 59: I don't think there is a `body` in the response:
https://github.com/mojaloop/sdk-scheme-adapter/blob/16ccfe52f101d38daa397cf3312f33462bf8176c/src/OutboundServer/handlers.js#L467-L475

kirgene (changes requested):

modules/api-svc/src/FSPIOPEventHandler/handlers/party-info-requested.js
```diff
@@ -56,7 +56,11 @@ module.exports.handlePartyInfoRequestedDmEvt = async (
...
+                partyResult: {
+                    party: response.party?.body,
```
kirgene, line 60: I don't think there is a `body` in the response:
https://github.com/mojaloop/sdk-scheme-adapter/blob/16ccfe52f101d38daa397cf3312f33462bf8176c/src/OutboundServer/handlers.js#L467-L475

kirgene (changes requested):

modules/api-svc/src/FSPIOPEventHandler/handlers/party-info-requested.js
vijayg10: Hi @kirgene , there are body and headers in party. Yest that's weird, but its the response from model currently. So I made this transformation as workaround

kirgene (approved):

modules/api-svc/src/FSPIOPEventHandler/handlers/party-info-requested.js
kleyow: Thats a carry over from merging the mojaloop-connector into the sdk-scheme-adapter. The `headers` were needed in the synchronous responses so now the schema is typically `objectResponse: {body:body, headers:headers}`.

kleyow: Confusing I know v - v but that was the way to keep everything standardized.

PR #377, opened 2022-09-15
feat: add test cases for fspiop and backend event servers

kirgene:
cover:
* POST /bulkTransactions
* PUT /bulkTransactions/{bulkTransactionId}
* PartyInfoRequested event
* BulkQuotesRequested event
* SDKOutboundBulkAcceptPartyInfoRequested event
* SDKOutboundBulkAcceptQuoteRequested event


modules/api-svc/src/OutboundServer/handlers.js
```diff
@@ -310,7 +310,7 @@ const putBulkTransactions = async (ctx) => {
             ctx.state.eventLogger.info(`Sent domain event ${msg.getName()}`);
         }
 
-        ctx.response.status = 204;
+        ctx.response.status = 202;
```
mdebarros, line 313: Can we use this enum for HTTP Status codes from central-services-shared? --> https://github.com/mojaloop/central-services-shared/blob/470f6e9a580b28d8def9d19519593d14bb990ecf/src/enums/http.js#L145

modules/api-svc/test/__mocks__/@mojaloop/sdk-scheme-adapter-private-shared-lib.js
```diff
@@ -0,0 +1,52 @@
+class KafkaDomainEventConsumer {
```
mdebarros, line 1: Let me know what you think about moving this to the private-shared-lib so it can be used for other modules?

mdebarros (commented):
Looks good, just some minor comments for you.

modules/api-svc/test/__mocks__/@mojaloop/sdk-scheme-adapter-private-shared-lib.js
kirgene: I think this mock is useful only within the api-sdk service. Other services are pretty simple and use DI everywhere.

mdebarros (approved):
+1

PR #384, opened 2022-09-22
fix(mojaloop/2939): missing lasterror in callback

vijayg10:
fix([mojaloop/2939](https://github.com/mojaloop/project/issues/2939)): missing lasterror in callback

Summary: For a bulkTransactions request, when the payee sends an ID not found error, that error should be populated in the lastError message field in the PUT /bulkTransactions/ID callback that gets sent to payer backend.

### DIFF
--- a/modules/api-svc/package.json
+++ b/modules/api-svc/package.json
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@koa/cors": "^3.4.1",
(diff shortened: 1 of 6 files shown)


mdebarros (approved):
+1

@vijayg10 be aware that the api-snippets is a breaking change. So I am not sure if you need to make any changes to accomodate the upgrade?

PR #386, opened 2022-09-22
fix(mojaloop/2941): handled party lookup timeout error

vijayg10:
fix([mojaloop/2941](https://github.com/mojaloop/project/issues/2941)): handled party lookup timeout error

Summary: When there is a timeout in the payer SDK, that error has to be sent back to payer sim backend via PUT/transactions/{transactionId} . But that message is not received at payer sim backend.


kirgene (approved):
LGTM

PR #390, opened 2022-09-23
fix(mojaloop/#2803): adjust transfer creation logic and int tests

kleyow:
- Add integration test for ProcessSDKOutboundBulkResponseSent 
- Add integration test for PrepareSDKOutboundBulkResponse
- Adjust bulk transfer request creation logic to exclude any batches that are not in the curroct state and quotes that had previously erroed.
- Adjust ProcessBulkTransfersCallback test.


modules/outbound-command-event-handler/src/domain/bulk_transaction_agg/index.ts
```diff
@@ -433,62 +434,65 @@ export class BulkTransactionAgg extends BaseAggregate<BulkTransactionEntity, Bul
...
-                        extensions: individualTransfer.request.quoteExtensions,
-                    };
-                    this._logger.debug(`Generated Transfers Request ${JSON.stringify(individualBulkTransferRequest, null, 2)} for BulkBatch.bulkQuoteId=${bulkBatch.bulkQuoteId}`);
-                    // Add Transfers to batch
-                    bulkBatch.addIndividualTransfer(
-                        individualBulkTransferRequest,
-                        individualTransfer.id,
-                    );
+            if(bulkBatch.state == BulkBatchInternalState.AGREEMENT_COMPLETED) {
+                const individualQuoteResults = bulkBatch.bulkQuotesResponse?.individualQuoteResults; // TODO: Are we not missing the condition here?
```
mdebarros, line 438:
```suggestion
                const individualQuoteResults = bulkBatch.bulkQuotesResponse?.individualQuoteResults;
```

TODO comment is no longer valid since we deal with the condition on line 476.

mdebarros (approved):
LGTM, just a slight update to remove a comment that is no longer applicable.

kleyow:
Follow up story to update libraries that necessitates shared library updates https://github.com/mojaloop/project/issues/2949

PR #394, opened 2022-10-03
feat(mojaloop/#2914): import api server with health endpoint

kleyow:
feat(mojaloop/#2914): import api server with health endpoint - https://github.com/mojaloop/project/issues/2914

- Port api server to domain handler
- Add key for repository connection in both handlers
- Cherrypicked commits from https://github.com/mojaloop/sdk-scheme-adapter/pull/393 to address CI building issues


modules/outbound-command-event-handler/src/api-server/models/health.ts
```diff
@@ -25,4 +25,5 @@
 export interface Health {
     status: 'OK' | 'ERROR';
     errors?: string[];
+    bulkTransactionRepoConnected?: boolean;
```
mdebarros, line 28: What about Kafka?

modules/outbound-domain-event-handler/src/api-server/handlers/health.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Context, Document } from 'openapi-backend';
+import Express from 'express';
+import { Health } from '../models';
+import { HealthService } from '../services/health';
+
+export default {
+    getHealth: async (_c: Context<Document>, req: Express.Request, res: Express.Response) => {
+        const health: Health = await new HealthService().getHealth();
+        const bulkTransactionRepo = req.app.get('bulkTransactionRepo');
+        health.bulkTransactionRepoConnected = bulkTransactionRepo.canCall();
```
mdebarros, line 34: What about Kafka?

And shouldn't this logic be in the `HealthService`?

modules/outbound-domain-event-handler/src/api-server/models/health.ts
```diff
@@ -0,0 +1,29 @@
...
+ - Name Surname <name.surname@gatesfoundation.com>
+ * Modusbox
+ - Kevin Leyow <kevin.leyow@modusbox.com>
+ --------------
+ ******/
+
+export interface Health {
+    status: 'OK' | 'ERROR';
+    errors?: string[];
+    bulkTransactionRepoConnected?: boolean;
```
mdebarros, line 28: Kafka?

modules/outbound-command-event-handler/src/api-server/handlers/health.ts
```diff
@@ -28,8 +28,10 @@ import { Health } from '../models';
 import { HealthService } from '../services/health';
 
 export default {
-    getHealth: async (_c: Context<Document>, _req: Express.Request, res: Express.Response) => {
+    getHealth: async (_c: Context<Document>, req: Express.Request, res: Express.Response) => {
         const health: Health = await new HealthService().getHealth();
+        const bulkTransactionRepo = req.app.get('bulkTransactionRepo');
+        health.bulkTransactionRepoConnected = bulkTransactionRepo.canCall();
```
mdebarros, line 34: Shouldn't this logic be in the `HealthService`?

modules/outbound-domain-event-handler/src/api-server/services/health.ts
```diff
@@ -0,0 +1,36 @@
...
+ - Kevin Leyow <kevin.leyow@modusbox.com>
+ --------------
+ ******/
+
+import { Health } from '../models';
+
+export class HealthService {
+    public async getHealth(): Promise<Health> {
+        const status: 'OK' | 'ERROR' = 'OK';
+        const errors: string[] = [];
```
mdebarros, line 30: This is where the additional checks should go for backends (i.e. BulkTransactionRepo and Kafka, etc).

modules/outbound-domain-event-handler/src/application/index.ts
```diff
@@ -67,6 +78,8 @@ import Config from '../shared/config';
         logger.info('\tDestroying outboundEventHandler handler...');
 
         await outboundEventHandler.destroy();
+        await bulkTransactionEntityRepo?.destroy();
+        await apiServer.stopServer();
```
mdebarros, line 82: We should really change the `stopServer` to `destroy` to align with this elsewhere throughout our code. <-- not something that we have to do right now, but something worth thinking about.

modules/outbound-domain-event-handler/src/shared/config.ts
```diff
@@ -43,6 +47,20 @@ const config = Convict({
         default: 'info',
         env: 'LOG_LEVEL',
     },
+    API_SERVER: {
+        ENABLED: {
+            doc: 'Whether to enable API server or not',
+            format: 'Boolean',
+            default: false,
+            env: 'API_SERVER_ENABLED',
```
mdebarros, line 55: I think we need to make these specific to the handler now, perhaps something like this:
```suggestion
            env: 'DOMAIN_EVENT_API_SERVER_ENABLED',
```

Note we should go back and update the Command handler to match the same as above, except using something like a `COMMAND_EVENT_*` prefix

modules/outbound-domain-event-handler/src/shared/config.ts
```diff
@@ -43,6 +47,20 @@ const config = Convict({
...
+        },
+        PORT: {
+            doc: 'The port of the API server.',
+            format: 'port',
+            default: 8001,
+            env: 'API_SERVER_PORT',
```
mdebarros, line 61: I think we need to make these specific to the handler now, perhaps something like this:

```suggestion
            env: 'DOMAIN_EVENT_API_SERVER_PORT',
```

Note we should go back and update the Command handler to match the same as above, except using something like a `COMMAND_EVENT_*` prefix

modules/outbound-domain-event-handler/src/shared/config.ts
mdebarros: Alternative perhaps the prefixes should be simplified to something like:
- `SDK_ODH_*` <-- ODH = outbound domain handler
- `SDK_OCH_*` <-- OCH = outbound command handler

etc.

This would only be applicable for configs that are specific to the Service/Handler, and where there is commonality, it could stay as is --- unless we can think of a reason not too do so?

modules/outbound-command-event-handler/src/api-server/models/health.ts
mdebarros, line 28: Kafka?

kleyow: I had checked @mojaloop/platform-shared-lib-nodejs-kafka-client-lib and couldn't find anything that was a clear cut health check.

I had planned to creating a story for that or try and figure out a faux health check.

modules/outbound-domain-event-handler/src/shared/config.ts
kleyow: Since the configs are entirely seperate no prefix should should be fine for all configs. We haven't added prefix for any other configs.

Or is there a case that we'd have both handlers running from the same enviorment, then we'd have to prefix everything.

mdebarros: > Or is there a case that we'd have both handlers running from the same enviorment, then we'd have to prefix everything.

We do, `yarn start` when we run the Integration Tests.

mdebarros (approved):
+1

PR #391, opened 2022-09-28
fix: bugs 2946, 2951

kirgene:
* fix #2946
* fix #2951

PR #395, opened 2022-10-04
fix: package run commands referencing incorrect module names

mdebarros:
- fixed run commands to call the correct module name for start run scripts on the root
- fixed start/server commands for dom/cmd handlers to not call tsc
- upgraded dependencies, except for the following which was added to 'ncurc.yaml ' ignore on dep:checks:
    - 'nx' - Upgrading this causes builds to fail (dependency relationship/graph are not being respected). TODO: Remove and upgrade once issue is resolved.
    - '@mojaloop/api-snippets' - This should be removed and upgraded as part of this PR: https://github.com/mojaloop/sdk-scheme-adapter/pull/391


kleyow (approved):

kleyow:
@mdebarros fixed int tests. going to merge this.

PR #402, opened 2022-10-18
feat(mojaloop#2965): end-end functional tests

sri-miriyala:
feat(mojaloop#2965): end-end functional tests https://github.com/mojaloop/project/issues/2965


modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -501,30 +501,37 @@ class InboundTransfersModel {
             const mojaloopResponse = shared.internalBulkQuotesResponseToMojaloop(response);
 
             // create our ILP packet and condition and tag them on to our internal quote response
+            this._logger.info(JSON.stringify({bulkQuoteRequest}));
```
mdebarros, line 504: I am not sure its a good idea to "stringify" a bulk message like this?

modules/api-svc/src/lib/model/lib/shared.js
```diff
@@ -428,6 +471,61 @@ const internalBulkQuotesResponseToMojaloop = (internal) => {
...
+            // mojaloop payee rejection and put the fsp backend error into an extension
+            // list.
+            if(transfer.errorResponse) {
+                externalTransfer = {
+                    transferId: transfer.transferId,
+                    errorInformation: Errors.MojaloopApiErrorObjectFromCode(Errors.MojaloopApiErrorCodes.PAYEE_REJECTION).errorInformation,
+                    extensionList: {
+                        extension: [
+                            {
+                                key: 'errorResponseStatusCode',
```
mdebarros, line 499: I assume this was a mapping that already existed?

modules/api-svc/src/lib/model/lib/shared.js
```diff
@@ -428,6 +471,61 @@ const internalBulkQuotesResponseToMojaloop = (internal) => {
...
+                                value: transfer.errorResponse.statusCode
+                            },
+                            {
+                                key: 'errorResponseMessage',
```
mdebarros, line 503: I assume this was a mapping that already existed?

mdebarros (commented):
Changes look more or less ok, but I have added a few comments above.

modules/api-svc/src/lib/model/lib/shared.js
kleyow: Not that I'm aware of. All the models that send a backend request to a dfsp are either expecting a successful response or some HTTP error. Theres no in between, the backend api makes little consideration for error messages. 

So in bulk where it's a successful response that contains nested quote/transfer errors I have no precedent of a mapping.

I've had lingering thoughts as to why the sdk doesn't just enforce that the backend it talks to use the FSPIOP api. We are putting all the mappings of a backend to the sdk in a flavoured `core-connector` for dfsp's anyway. 

It's hard to design backend api for a non-existent generic fsp.

modules/api-svc/src/lib/model/InboundTransfersModel.js
kleyow: Ah these stringifyies were left overs from debuging. They can be removed.

modules/api-svc/src/lib/model/lib/shared.js
mdebarros: We already have something like this here --> https://github.com/mojaloop/central-services-error-handling/blob/master/src/factory.js#L147

kleyow: Addressed in https://github.com/mojaloop/sdk-scheme-adapter/pull/404

modules/api-svc/src/lib/model/lib/shared.js
kleyow: Added in https://github.com/mojaloop/sdk-scheme-adapter/pull/404

modules/api-svc/src/lib/model/InboundTransfersModel.js
kleyow: Removed in https://github.com/mojaloop/sdk-scheme-adapter/pull/404

kleyow (approved):

PR #403, opened 2022-10-18
fix: added optional chaining to bulk transfer handler

kleyow:


kleyow:
@sri-miriyala this should fix the issue. the assertions only need to be updated.

mdebarros (approved):
+1

PR #411, opened 2022-11-03
fix: correct internal/external bulk quote convert function

kleyow:
- Fix issue with external to internal mapping.


sri-miriyala (dismissed):
LGTM

sri-miriyala:
👍

PR #412, opened 2022-11-04
chore: remove max items from put bulkTransactions

kleyow:
- Remove maxItems on component used in put /bulkTransactions individualTransferResults


mdebarros (approved):
+1

PR #413, opened 2022-11-04
feat: added two payee sdks and ttk as hub in func tests

vijayg10:


modules/api-svc/src/InboundServer/middlewares.js
```diff
@@ -356,9 +356,10 @@ const createLogger = (logger) => async (ctx, next) => {
         path: ctx.path,
         method: ctx.method
     }});
-    if (!ctx.state.logExcludePaths.includes(ctx.path)) {
-        ctx.state.logger.push({body: ctx.request.body}).log('Request received');
-    }
+    await ctx.state.logger.log('Request received');
+    // if (!ctx.state.logExcludePaths.includes(ctx.path)) {
```
mdebarros, line 360: I assume we dont have an option to enable debug, etc on the SDK right?

If not, we should add a TODO here, and even possibly re-enable this.

mdebarros (commented):
LGTM, just a single comment.

Let me know your thoughts.

mdebarros (approved):
+1

PR #440, opened 2023-02-27
fix(mojaloop/3137): bulk transactions sdk crash

vijayg10:
fix([mojaloop/3137](https://github.com/mojaloop/project/issues/3137)): bulk transactions sdk crash


mdebarros (approved):
+1

PR #447, opened 2023-05-26
feat(mojaloop/3347): enhance requesttopaytransfer

vijayg10:
feat([mojaloop/3347](https://github.com/mojaloop/project/issues/3347)): enhance requesttopaytransfer

### DIFF
--- a/docker/ml-testing-toolkit/spec_files/api_definitions/mojaloop_sdk_outbound_scheme_adapter_1.0/api_spec.yaml
+++ b/docker/ml-testing-toolkit/spec_files/api_definitions/mojaloop_sdk_outbound_scheme_adapter_1.0/api_spec.yaml
@@ -383,15 +383,15 @@ paths:
           $ref: '#/components/responses/transferServerError'
         '504':
           $ref: '#/components/responses/transferTimeout'
-  /requestToPayTransfer/{requestToPayTransactionId}:
+  /requestToPayTransfer/{transactionRequestId}:
     put:
       summary: >-
         Continues a transfer that has paused at the otp stage in order to accept
         or reject quote
       description: >
-        The HTTP request `PUT /transfers/{transferId}` is used to continue a
-        transfer initiated via the `POST /transfers` method that has halted
-        after party lookup and/or quotation stage.
+        This request is used to continue a requestToPayTransfer initiated via
+        the `POST /requestToPayTransfer` method that has halted after quotation
+        stage and/or otp stage.
 
 
         The request body should contain either the "acceptOTP" or "acceptQuote"
@@ -410,7 +410,7 @@ paths:
                 - $ref: '#/components/schemas/transferContinuationAcceptQuote'
                 - $ref: '#/components/schemas/transferContinuationAcceptOTP'
       parameters:
-        - $ref: '#/components/parameters/requestToPayTransactionId'
+        - $ref: '#/components/parameters/transactionRequestId'
       responses:
         '200':
           $ref: '#/components/responses/transferSuccess'
@@ -2370,7 +2370,8 @@ components:
     requestToPayTransferRequest:
       type: object
       required:
-        - requestToPayTransactionId
+        - homeR2PTransactionId
+        - transactionRequestId
         - from
         - to
         - amountType
@@ -2380,11 +2381,13 @@ components:
         - initiator
         - initiatorType
       properties:
-        requestToPayTransactionId:
+        homeR2PTransactionId:
           type: string
           description: >-
             Transaction ID from the DFSP backend, used to reconcile transactions
             between the Switch and DFSP backend systems.
+        transactionRequestId:
+          $ref: '#/components/schemas/CorrelationId'
         from:
           $ref: '#/components/schemas/transferParty'
         to:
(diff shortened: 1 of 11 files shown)


docker/ml-testing-toolkit/spec_files/api_definitions/fspiop_1.1/api_spec.yaml
```diff
@@ -1198,6 +1198,27 @@ paths:
       parameters:
         #Headers
         - $ref: "#/components/parameters/Accept"
+        ## Adding the following query parameters for testing as TTK is enforcing query params validation which should be fixed in TTK
```
mdebarros, line 1201: Why were we missing these?

I.e. is this missing from the FSPIOP swagger?

docker/ml-testing-toolkit/spec_files/api_definitions/mojaloop_sdk_outbound_scheme_adapter_1.0/api_spec.yaml
```diff
@@ -383,15 +383,15 @@ paths:
...
+  /requestToPayTransfer/{transactionRequestId}:
     put:
       summary: >-
         Continues a transfer that has paused at the otp stage in order to accept
         or reject quote
       description: >
-        The HTTP request `PUT /transfers/{transferId}` is used to continue a
-        transfer initiated via the `POST /transfers` method that has halted
-        after party lookup and/or quotation stage.
+        This request is used to continue a requestToPayTransfer initiated via
```
mdebarros, line 392: Please can you update this swagger. 

I see some of the suggested changes in the API Snippets PR did not propagate here.

modules/api-svc/.env.example
```diff
@@ -93,6 +93,17 @@ AUTO_ACCEPT_PARTY=false
 # confirmation call will be required to progress the requestToPay to transactionRequest state.
 AUTO_ACCEPT_R2P_PARTY=false
 
+# This parameter is only for the requestToPay transfers when the initiator is of type BUSINESS.
+# if set to false the SDK will not automatically accept quote on a requestToPayTransfer.
+# but will halt the operation after a quote response is received. A further
+# confirmation call will be required to progress the requestToPayTransfer to authorization step.
+AUTO_ACCEPT_R2P_BUSINESS_QUOTES=false
+
+# if set to false the SDK will not automatically accept OTP on a requestToPay operation
```
mdebarros, line 102: I don't think anything needs to change right now, but I think we need to relook at this config to see if it can be more generic, and not specific to OTP.

modules/api-svc/src/lib/model/OutboundRequestToPayTransferModel.js
```diff
@@ -212,6 +229,10 @@ class OutboundRequestToPayTransferModel {
                 // request an OTP
                 return this._requestOTP();
 
+            case 'skipOTP':
+                // request an OTP
```
mdebarros, line 233:
```suggestion
                // Skip OTP request
```

modules/api-svc/src/lib/model/OutboundRequestToPayTransferModel.js
```diff
@@ -231,12 +252,12 @@ class OutboundRequestToPayTransferModel {
      * sent because the OTP did not match.
      */
     async rejectRequestToPay() {
-        const authResponse = {
-            responseType: 'REJECTED'
+        const mojaloopResponse = {
+            transactionRequestState: 'REJECTED'
```
mdebarros, line 256: It would be so nice to use enums instead of retyping out the status/states explicitly. 

Perhaps a new story is needed to address that.

test/func/ttk-tests-docker-compose.yml
```diff
@@ -16,8 +16,8 @@ services:
...
-        npm run cli -- -u http://ttk-ttksim1:5050 -l 2 -i /tmp/test_cases/testing-toolkit-test-cases-feat-3344-EPIC-Enhance-SDK-Scheme-Adaptor-to-support-the-request-to-Pay-use-case/collections/hub/sdk_scheme_adapter/request-to-pay/basic -e environments/hub_local_environment.json --report-format html --report-target file://reports/ttk-func-r2p-tests-report.html
+        npm run cli -- -u http://ttk-ttksim1:5050 -l 2 -i /tmp/test_cases/testing-toolkit-test-cases-15.0.0-sdk.1/collections/hub/sdk_scheme_adapter/bulk/basic -e environments/hub_local_environment.json --report-format html --report-target file://reports/ttk-func-bulk-tests-report.html &&
+        npm run cli -- -u http://ttk-ttksim1:5050 -l 2 -i /tmp/test_cases/testing-toolkit-test-cases-15.0.0-sdk.1/collections/hub/sdk_scheme_adapter/request-to-pay/basic -e environments/hub_local_environment.json --report-format html --report-target file://reports/ttk-func-r2p-tests-report.html
```
mdebarros, line 23: Would it make sense to change `15.0.0-sdk.1` to reference an environment variable instead, i.e. we change it in one place and it changes everywhere?

mdebarros (commented):
Looks good to me.

Just some minor comments for your attention.

docker/ml-testing-toolkit/spec_files/api_definitions/fspiop_1.1/api_spec.yaml
vijayg10: Yes these query parameters are missing from FSPIOP spec. I am not sure if that's a bug or intentional. But TTK is not accepting the query params without having them in spec file. So I added them for testing with a comment.

modules/api-svc/.env.example
vijayg10: I didn't change anything. I just added those existing config params to example env file.

docker/ml-testing-toolkit/spec_files/api_definitions/fspiop_1.1/api_spec.yaml
mdebarros: https://github.com/mojaloop/mojaloop-specification/blob/master/fspiop-api/documents/v1.1-document-set/fspiop-v1.1-openapi3.yaml#L1159 <-- interesting that its missing.

PR #448, opened 2023-06-06
fix(mojaloop/3322): put transactionrequests error handler

vijayg10:
fix([mojaloop/3322](https://github.com/mojaloop/project/issues/3322)): put transactionrequests error handler


mdebarros (approved):
+1

PR #458, opened 2023-11-24
feat(3618): added support for POST /fxQuotes and /fxTransfers inbound requests

geka-evk:
feat(3618): added support for POST /fxQuotes and /fxTransfers inbound requests


aaronreynoza (approved):

modules/api-svc/src/InboundServer/handlers.js
```diff
@@ -937,12 +950,39 @@ const putBulkTransfersByIdError = async(ctx) => {
...
 };
 
+const postFxQuotes = async (ctx) => {
+    const sourceFspId = ctx.request.headers['fspiop-source'];
+    const { body, headers } = ctx.request;
+    const { logger } = ctx.state;
+    const logPrefix = 'Handling POST fxQuotes request';
+
+    const model = createInboundTransfersModel(ctx);
+    model.postFxQuotes({ body, headers }, sourceFspId)
```
vijayg10, line 965: You need to shallow copy the body and headers before calling the async function here. This is to avoid parallel processing issue we got previously on high concurrent calls. Please refer the functions for quotes and transfers.

modules/api-svc/src/InboundServer/handlers.js
```diff
@@ -937,12 +950,39 @@ const putBulkTransfersByIdError = async(ctx) => {
...
+};
+
+const postFxTransfers = async (ctx) => {
+    const sourceFspId = ctx.request.headers['fspiop-source'];
+    const { body, headers } = ctx.request;
+    const { logger } = ctx.state;
+    const logPrefix = 'Handling POST fxTransfers request';
+
+    const model = createInboundTransfersModel(ctx);
+    model.postFxTransfers({ body, headers }, sourceFspId)
```
vijayg10, line 979: You need to shallow copy the body and headers before calling the async function here. This is to avoid parallel processing issue we got previously on high concurrent calls. Please refer the functions for quotes and transfers.

modules/api-svc/src/lib/dto.js
```diff
@@ -0,0 +1,29 @@
...
+        body: request.body
+    },
+    currentState: SDKStateEnum.QUOTE_REQUEST_RECEIVED,
+    initiatedTimestamp: new Date().toISOString(),
+});
+
+const fxQuoteRequestStateDto = (request) => ({
+    // assume commitRequestId from fxTransfer should be same as conversionTerms.conversionId from fxQuotes
+    conversionId: request.body.conversionTerms.conversionId,
+    fxQuoteRequest: request,
```
vijayg10, line 20: Is it ok to pass the whole request object here instead of following quote implementation? I think its better to follow the same (passing only headers and body).

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -77,7 +75,9 @@ class InboundTransfersModel {
     updateStateWithError(err) {
         this.data.lastError = err;
         this.data.currentState = SDKStateEnum.ERROR_OCCURRED;
-        return this._save();
+        return this.data.transferId // if no transferId - we are in fxQuote flow
```
vijayg10, line 78: Why do we need this condition?

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -875,15 +978,18 @@ class InboundTransfersModel {
...
+        if (this.data) {
             //we have persisted state so update that with this error
             this.data.lastError = {
                 originalError: err.stack || util.inspect(err),
-                mojaloopError: mojaloopError,
+                mojaloopError,
             };
             this.data.currentState = SDKStateEnum.ERROR_OCCURRED;
-            await this._save();
+            this.data.transferId
```
vijayg10, line 988: Do we need this condition?

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -916,7 +1022,30 @@ class InboundTransfersModel {
             throw err;
         }
     }
+
+    // todo: combine with this._save
+    async saveFxState() { // fxQuote + fxTransfer
```
vijayg10, line 1027: Do we need a separate function for fxState?

modules/api-svc/src/lib/model/common/Enums.js
```diff
@@ -17,8 +17,11 @@ const SDKStateEnum = {
     WAITING_FOR_ACTION: 'WAITING_FOR_ACTION',
     WAITING_FOR_PARTY_ACCEPTANCE: 'WAITING_FOR_PARTY_ACCEPTANCE',
     WAITING_FOR_AUTH_ACCEPTANCE: 'WAITING_FOR_AUTH_ACCEPTANCE',
-    QUOTE_REQUEST_RECEIVED: 'QUOTE_REQUEST_RECEIVED',
     WAITING_FOR_QUOTE_ACCEPTANCE: 'WAITING_FOR_QUOTE_ACCEPTANCE',
+    QUOTE_REQUEST_RECEIVED: 'QUOTE_REQUEST_RECEIVED',
+    FX_QUOTE_REQUEST_RECEIVED: 'FX_QUOTE_REQUEST_RECEIVED',
+    FX_QUOTE_WAITING_FOR_ACCEPTANCE: 'FX_QUOTE_WAITING_FOR_ACCEPTANCE',
```
vijayg10, line 23: Are we using this state 'FX_QUOTE_WAITING_FOR_ACCEPTANCE'?

modules/api-svc/src/lib/model/common/Enums.js
```diff
@@ -33,7 +36,18 @@ const TransactionRequestStateEnum = {
     REJECTED: 'REJECTED',
 };
 
+const Directions = Object.freeze({
+    INBOUND: 'INBOUND',
+    OUTBOUND: 'OUTBOUND',
+});
+
+const CacheKeyPrefixes = Object.freeze({
+    FX_QUOTE: 'fxQuote_in'
```
vijayg10, line 45: Do we need a separate prefix for FX cache? Lets discuss this

modules/api-svc/test/unit/api/accounts/accounts.test.js
```diff
@@ -114,7 +114,9 @@ describe('Outbound Accounts API', () => {
                 postAccountsErrorTimeoutResponse);
         });
 
-        test('should return mojaloop error response', () => {
+        // todo: fix this test, coz it doesn't work with api.yaml v.2
+        //       After the test will be fixed, we need remove /InboundServer/api_v1.yaml
+        test.skip('should return mojaloop error response', () => {
```
vijayg10, line 119: I suggest not to skip the test. Let CI check fail.

vijayg10 (changes requested):

modules/api-svc/src/lib/model/InboundTransfersModel.js
geka-evk: `_save()` uses different cache key based on _data.transferId_:
https://github.com/mojaloop/sdk-scheme-adapter/blob/feat/fx-impl-3618/modules/api-svc/src/lib/model/InboundTransfersModel.js#L1001

in FX payload we don't have _transferId f_ield

modules/api-svc/src/InboundServer/handlers.js
geka-evk: Could please, give a bit more context about issue you mentioned. For each concurrent calls we have a separate `ctx`, thus `body` and `headers` are also completely new for each concurrent calls. Sorry, but I'm not sure I get the reason, why shallow copy is needed.

modules/api-svc/src/lib/dto.js
geka-evk: but `request` contains only _body_ and _headers_  (first param to _model.postFxQuotes()_ method):
https://github.com/mojaloop/sdk-scheme-adapter/blob/feat/fx-impl-3618/modules/api-svc/src/InboundServer/handlers.js#L965

modules/api-svc/src/lib/model/common/Enums.js
geka-evk: I thought, we should save _quotes_ and _fxQuotes_ with different keys (coz they are different "entities"). But I might be wrong. Let's discuss

modules/api-svc/src/InboundServer/handlers.js
vijayg10: We ran into this situation with “Koa” API library we are using. This library passes a mutated context object instead of a fresh one and as a result if we pass request object by reference it will get mutated by the time async function make use of the contents of it. That's why we are shallow copying the body and headers and pass a fresh object reference. We literally got this issue when we handled high load and figured out the issue with this and this approach solved the issue. Here is a blog post we created based on our findings at that time https://www.infitx.com/async-operations-in-javascript/

modules/api-svc/src/lib/model/common/Enums.js
geka-evk: here's:
https://github.com/mojaloop/sdk-scheme-adapter/blob/feat/fx-impl-3618/modules/api-svc/src/lib/model/InboundTransfersModel.js#L547

modules/api-svc/src/lib/model/InboundTransfersModel.js
geka-evk: to save state either for _quote_ or for _fxQuote_

modules/api-svc/src/InboundServer/handlers.js
vijayg10: This is the old PR for this fix https://github.com/mojaloop/sdk-scheme-adapter/pull/440

modules/api-svc/src/lib/model/InboundTransfersModel.js
geka-evk: to save state either for _quote_ or for _fxQuote_,   (coz for now these methods have a slightly different impl. (in particular, different _cache keys_)

modules/api-svc/test/unit/api/accounts/accounts.test.js
geka-evk: But if unit-tests fail, will I be able push my commits and will integration and functional tests be run?

modules/api-svc/src/InboundServer/handlers.js
geka-evk: I'm not sure it's related to Koa API, coz we don't use _contex_t object in our models:
https://github.com/mojaloop/sdk-scheme-adapter/blob/feat/fx-impl-3618/modules/api-svc/src/InboundServer/handlers.js#L970
We extract needed data, and models deal only with _body, headers, sourceFspId, ..._ as with separate data (not a part of context)
Anyway, I did what you asked, please, check

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -515,7 +515,17 @@ class InboundTransfersModel {
             await this.saveFxState();
 
             const internalRequest = shared.mojaloopFxQuoteRequestToInternal(body);
-
+            
+            // Check if transactionRequestId exists in cache 
+            if(body.transactionRequestId) {
```
geka-evk: but we don't have `transactionRequestId` in _POST /fxQuotes_ payload according Swagger docs

modules/api-svc/src/lib/model/common/Enums.js
vijayg10: As we have a condition and calling different functions for save to cache, lets use separate keys for now.

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -521,6 +508,132 @@ class InboundTransfersModel {
...
+            this.data = dto.fxQuoteRequestStateDto(request);
+            await this.saveFxState();
+
+            const internalRequest = shared.mojaloopFxQuoteRequestToInternal(body);
+            
+            // Check if conversionId exists in cache 
+            if(body.conversionId) {
+                const previousTxnReq = await this.loadFxState(body.conversionId);
+                if(previousTxnReq) {
+                    internalRequest.homeR2PconversionId = previousTxnReq.homeR2PconversionId;
```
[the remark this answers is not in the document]
vijayg10: As we discussed, lets comment this for now and merge the PR to feature branch. We can discuss it with @aaronreynoza later

vijayg10 (approved):

PR #459, opened 2023-11-28
feat(mojaloop/#3655): add support for parties supportedcurrencies & kyc

aaronreynoza:


modules/api-svc/src/config.js
```diff
@@ -65,6 +65,7 @@ module.exports = {
     logLevel: env.get('LOG_LEVEL').default('info').asEnum(['trace', 'debug', 'info', 'warn', 'error', 'fatal']),
     inbound: {
         port: env.get('INBOUND_LISTEN_PORT').default('4000').asPortNumber(),
+        supportedCurrencies: (env.get('SUPPORTED_CURRENCIES').asString()).split(','),
```
geka-evk, line 68: You can use `.asArray()` function to convert CSV string to an array:
```
supportedCurrencies: env.get('SUPPORTED_CURRENCIES').asArray,
```
https://github.com/evanshortiss/env-var/blob/HEAD/API.md#asarraydelimiter-string

vijayg10: Done

PR #465, opened 2024-02-05
feat(mojaloop/#3670): add standard components lib update for k6 validation 

aaronreynoza:
feat(mojaloop/#3670): add standard components lib update for k6 validation


modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -156,6 +156,14 @@ class InboundTransfersModel {
                 party: shared.internalPartyToMojaloopParty(response, this._dfspId)
             };
 
+            let { tracestate = undefined, traceparent = undefined } = headers;
```
kleyow, line 159: I'm not familar with type of assignment, can you explain?

aaronreynoza: So headers in this scenario is the object with both tracestate and traceparent. Both are optional and required only for the Perf tests. so if they're not present they'll be undefined, as a regular request. the values for both variables are later validated to complete the tracestate callback, but not if they're not sent

PR #513, opened 2024-11-11
feat(csi-128): updated sdk-standard-components with axios impl.; fixed unit-tests

geka-evk:
1. Updated _sdk-standard-components_ with **axios** as http client;
2. Added retry mechanism
3. Improved logging to follow context-logger output format
image


modules/api-svc/package.json
```diff
@@ -67,12 +67,12 @@
...
     "@mojaloop/central-services-logger": "^11.5.1",
-    "@mojaloop/central-services-metrics": "^12.0.8",
-    "@mojaloop/central-services-shared": "18.11.0",
+    "@mojaloop/central-services-metrics": "^12.1.0",
+    "@mojaloop/central-services-shared": "18.11.1",
     "@mojaloop/event-sdk": "^14.1.1",
     "@mojaloop/ml-schema-transformer-lib": "2.3.4",
     "@mojaloop/sdk-scheme-adapter-private-shared-lib": "workspace:^",
-    "@mojaloop/sdk-standard-components": "^19.4.1",
+    "@mojaloop/sdk-standard-components": "19.5.0-snapshot.12",
```
kalinkrustev, line 75: putting here a reminder to merge @mojaloop/sdk-standard-components first

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-scheme-adapter",
-  "version": "23.6.0-iso.3",
+  "version": "23.7.0-snapshot.1",
```
kalinkrustev, line 3: this must be restored to 23.6.0-iso.3 before merging

kalinkrustev (changes requested):
I do not know very well the SDK, leaving some generic comments

package.json
geka-evk: Shouldn't it be done automatically after merging into `minor/iso`?

kalinkrustev: Yes, you are right, but then we will need to update the CD to track the new range
See https://github.com/infitx-org/profile-cd/blob/main/updatecli.d/sdk-scheme-adapter.yaml

modules/api-svc/package.json
geka-evk: done

package.json
geka-evk: done

PR #524, opened 2024-12-20
feat(ime-392): disable iso cache

Ujjwal-Izyane:
- Fixes the issue for experience api crashing because of the iso cache
- Transforms iso messages to fspiop before caching
- Save the cache in a form that works with experience api
- Remove log statements added for debugging the ime/306


modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -258,7 +258,7 @@ class InboundTransfersModel {
             }
             this.data.quoteResponse = {
                 headers: res.originalRequest.headers,
-                body: res.originalRequest.body,
+                body: JSON.stringify(mojaloopResponse),
```
vijayg10, line 261: Why do we need JSON stringify? I think its better to pass object directly.

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -466,9 +466,10 @@ class InboundTransfersModel {
             // make a callback to the source fsp with the transfer fulfilment
             const res = await this._mojaloopRequests.putTransfers(prepareRequest.transferId, mojaloopResponse,
                 sourceFspId);
+
             this.data.fulfil = {
                 headers: res.originalRequest.headers,
-                body: res.originalRequest.body,
+                body: JSON.stringify(mojaloopResponse),
```
vijayg10, line 472: Why do we need JSON stringify? I think its better to pass object directly.

modules/api-svc/src/lib/model/InboundTransfersModel.js
```diff
@@ -573,12 +574,8 @@ class InboundTransfersModel {
 
             this.data.fxQuoteResponse = {
                 headers: res.originalRequest.headers,
-                body: res.originalRequest.body,
+                body: JSON.stringify(mojaloopResponse),
```
vijayg10, line 577: Why do we need JSON stringify? I think its better to pass object directly.

modules/api-svc/src/lib/model/OutboundTransfersModel.js
```diff
@@ -560,8 +560,14 @@ class OutboundTransfersModel {
             const subscribing = this._cache.subscribeToOneMessageWithTimer(channel);
 
             const resp = await this._requests.postFxQuotes(payload, payload.conversionTerms.counterPartyFsp);
+            
             const { originalRequest } = resp;
-            this.data.fxQuoteRequest = originalRequest;
+            // Setting the fxQuoteRequest to have the fspiop payload 
+            // If ISO20022 is required then use originalRequest
+            this.data.fxQuoteRequest = {
+                body: JSON.stringify(payload),
```
vijayg10, line 568: Why do we need JSON stringify? I think its better to pass object directly.

modules/api-svc/src/lib/model/InboundTransfersModel.js
Ujjwal-Izyane: The experience api is expecting these to be in string format , and if they are not a string it will throw and exception.
That's why this is converted to a string before it is saved to the redis cache.
The case in which it is an object is not handled properly, that will require some minor changes to experience api.

modules/api-svc/src/lib/model/InboundTransfersModel.js
Ujjwal-Izyane: The same as above , won't work with the current code of the experience api.

modules/api-svc/src/lib/model/InboundTransfersModel.js
vijayg10: My worry is JSON.stringify is very demanding opertation. I think already the whole data object is being converted to string when storing into redis. This is on top of it. We can adjust experience api, sdk-scheme-adapter is the critical component in transfers flow.

Ujjwal-Izyane: Okay ,let's do that

PR #555, opened 2025-03-05
feat: add code to restart sdk on broken control client connection

kleyow:


modules/api-svc/src/index.js
```diff
@@ -284,12 +308,33 @@ class Server extends EventEmitter {
...
-                    this.logger.push({ newConf }).debug('Control client closed. Restarting server...');
-                    this.restart(_.merge({}, newConf, {
-                        control: { stopped: Date.now() }
-                    }));
-                }, RESTART_INTERVAL_MS));
+
+                const schedulePing = () => {
+                    clearTimeout(this.pingTimeout);
+                    this.pingTimeout = setTimeout(() => {
+                        this.logger.push({ newConf }).error('Ping timeout, possible broken connection. Restarting server...');
```
geka-evk, line 315: Are we sure we want to log  the whole `newConf` object with `error` severity? It might have sensitive data (e.g. private keys) and be too long

PR #601, opened 2025-08-05
fix: concurrency issues

vijayg10:


modules/api-svc/src/lib/cache.js
```diff
@@ -196,6 +199,109 @@ class Cache {
...
+            let timeoutHandle = null;
+            let subscription = null;
+            let isResolved = false;
+
+            // Helper to safely unsubscribe from Redis channel
+            const unsubscribeFromRedis = async (reason = 'cleanup') => {
+                if (this._channelEmitter.listenerCount(channel) === 0 && this._subscriptionClient) {
+                    try {
+                        await this._subscriptionClient.unsubscribe(channel);
+                        this._logger.isDebugEnabled && this._logger.push({ channel, reason }).debug('Unsubscribed from Redis channel');
```
geka-evk, line 232: I believe, we don't need to check `this._logger.isDebugEnabled &&` coz  `logger.debug` method already has that check

modules/api-svc/src/lib/cache.js
```diff
@@ -196,6 +199,109 @@ class Cache {
...
+                    } catch (unsubscribeErr) {
+                        this._logger.isWarnEnabled && this._logger.push({ channel, reason, err: unsubscribeErr }).warn('Failed to unsubscribe from Redis channel');
```
geka-evk, line 234: we'll lose error info in logs with such approach:
Image

I'd suggest always adding error as a second argument to logger methods, e.g.:
```
this._logger.push({ channel, reason }).warn('Failed to unsubscribe from Redis channel: ', unsubscribeErr);
```

modules/api-svc/src/lib/model/OutboundTransfersModel.js
```diff
@@ -321,147 +322,121 @@ class OutboundTransfersModel {
...
-                // cancel the timeout and unsubscribe before rejecting the promise
-                clearTimeout(timeout);
-
-                // we dont really care if the unsubscribe fails but we should log it regardless
-                this._cache.unsubscribe(payeeKey, subId, true).catch(e => {
-                    this._logger.isErrorEnabled && this._logger.error(`Error unsubscribing ${payeeKey} ${subId}: ${e.stack || safeStringify(e)}`);
-                });
-
-                return reject(err);
+                this._logger.isErrorEnabled && this._logger.error(`Error in resolvePayee ${payeeKey}: ${err.stack || safeStringify(err)}`);
```
geka-evk, line 424: it's better to pass `err` as a 2nd argument to logger:
```
this._logger.error(`Error in resolvePayee ${payeeKey}: `, err);
```

geka-evk (dismissed):

modules/api-svc/src/lib/model/OutboundTransfersModel.js
vijayg10: Thanks, done

modules/api-svc/src/lib/cache.js
vijayg10: Thanks, done

geka-evk (approved):

modules/api-svc/src/lib/model/OutboundTransfersModel.js
```diff
@@ -421,7 +421,7 @@ class OutboundTransfersModel {
                 return resolve(payee);
             }
             catch(err) {
-                this._logger.isErrorEnabled && this._logger.error(`Error in resolvePayee ${payeeKey}: ${err.stack || safeStringify(err)}`);
+                this._logger.error(`Error in resolvePayee ${payeeKey}:`, err.stack || safeStringify(err));
```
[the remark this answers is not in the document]
vijayg10: Done

PR #676, opened 2026-06-02
fix: make maxsocket configurable so it doesnt use infinite number of sockets #4480

shashi165:
Fixes unbounded TLS connection growth under load by capping the outbound mTLS `https.Agent` to a configurable `maxSockets` limit and sharing a single agent instance across all model instantiations instead of creating a new one (with `maxSockets: Infinity`) per request.

added an env variable `OUTBOUND_MAX_SOCKETS` to control the maxSockets with default value of 100


elnyry-sam-k (dismissed):
👍

modules/api-svc/src/config.js
```diff
@@ -107,6 +107,7 @@ module.exports = {
     },
     outbound: {
         port: env.get('OUTBOUND_LISTEN_PORT').default('4001').asPortNumber(),
+        maxSockets: env.get('OUTBOUND_MAX_SOCKETS').default('100').asInt(),
```
elnyry-sam-k, line 110: in both sdkserver and index, 256 is set as default; maybe remove it there? 

> maxSockets: conf.outbound?.maxSockets || 256,

shashi165: removed redundant 256

elnyry-sam-k (dismissed):
👍

elnyry-sam-k (approved):
👍

PR #678, opened 2026-06-05
chore: fix 3586 nx upgrade

gibaros:
Address https://github.com/mojaloop/project/issues/3586


gibaros:
Team decided to go another route and instead of upgrading NX as performed on this PR, team decided to migrate from NX to Yarn 4 as the size and complexity of this monorepo does not warrants NX.

Yarn 4 migration merged at: https://github.com/mojaloop/sdk-scheme-adapter/pull/681

PR #680, opened 2026-07-23
fix: iso transfers missing mandatory field

shashi165:
## Problem

Transfers sent via the outbound `/simpleTransfers`
endpoint are missing required ISO20022 fields when `API_TYPE=iso20022`, causing the
switch to reject them with FSPIOP error `3102` `"Missing mandatory element - ChrgBr"`.

## AI Assistance

- [ ] No AI tools were used
- [x] AI was used and I reviewed all generated code
- Tools: Claude Code Sonnet 5


gibaros (approved):
👍

